PR #6, opened 2019-02-28
Added Procfile

RuiAAPeres:
### Why?

The `Procfile` defines what commands to be executed when the app starts up in Heroku. For now, I will use what we previously had, although this states something else. https://github.com/vapor/vapor/issues/1591

I have also done some work from the Heroku side and was able to deploy it successfully:

screenshot 2019-02-28 at 22 23 28

### How?

- By adding the Procfile. 

### Are these changes (unit/automation/snapshot) tested?
N/A

PR checklist:

* [x] I've assigned this PR to myself
* [x] I've set all the necessary PR's labels

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,1 @@
+web: App --env=debug --workdir="./"


dmcrodrigues (approved):
LGTM

PR #7, opened 2019-04-13
Introduce `GitHubService` to handle incoming events from GitHub's webhook

dmcrodrigues:
`GitHubService` will work as a central entity to handle all the incoming events from GitHub's Webhooks. This events can then be observed by any interested parties to act upon them when they occur, in this case any pull request events will be monitored to trigger the necessary actions by the bot. This monitoring will be introduced in one of the following pull requests.

This also enforces a separation between the core of the bot and the network stack, in this case Vapor, to simplify upgrades or even migrate to a different stack if necessary. It also allows embed the bot in other infrastructures if necessary, e.g. extend an existent and custom bot to introduce this functionality in particular.

> ⚠️ NOTE: Securing the origin of any incoming payload will be introduced in the next pull request.


Package.swift
```diff
@@ -2,14 +2,17 @@
 import PackageDescription
 
 let package = Package(
-    name: "WallE",
+    name: "WALL·E",
```
dmcrodrigues: @RuiAAPeres let me know if you agree with this rename, if so we could rename the repository as well.

Package.resolved
```diff
@@ -55,6 +55,42 @@
           "version": "3.0.3"
         }
       },
+      {
+        "package": "Nimble",
```
AliSoftware, line 59: Is the addition of Nimble/Quick as new dependencies really necessary? Given that we are considering removing them for our other repos, might as well consider not adding it in here at all and use `XCTest` directly instead?

Sources/App/Extensions/Vapor.Request+RequestProtocol.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+import Bot
+import Vapor
+
+extension Vapor.Request: RequestProtocol {
+    public func header(named name: String) -> String? {
+        return http.headers["X-GitHub-Event"].first
```
AliSoftware: The `name` parameter isn't used there, and instead `X-GitHub-Event` is hardcoded, seems like an oversight?

```suggestion
        return http.headers[name].first
```

Sources/Bot/Services/GitHubService.swift
```diff
@@ -0,0 +1,45 @@
...
+
+public final class GitHubService {
+    private let eventsObserver: Signal<Event, NoError>.Observer
+    public let events: Signal<Event, NoError>
+
+    public init() {
+        (events, eventsObserver) = Signal.pipe()
+    }
+
+    public func handleEvent(from request: RequestProtocol) -> Result<Void, EventHandlingError> {
```
AliSoftware: Given that the return value is synchronous, wondering if it wouldn't be simpler to return a `EventHandlingError?` rather than `Result<Void, EventHandlingError>` here?

Sources/Bot/Services/GitHubService.swift
```diff
@@ -0,0 +1,45 @@
...
+
+    private func parseEvent(from request: RequestProtocol) -> Result<Event, EventHandlingError> {
+        guard let event = request.header(named: "X-GitHub-Event")
+            else { return .failure(.invalid) }
+
+        switch event {
+        case "pull_request":
+            return Result(request.decodeBody(PullRequestEvent.self), failWith: .invalid)
+                .map(Event.pullRequest)
+        case "ping":
```
AliSoftware: Might be worth creating some constants for `X-GitHub-Event`, `pull_request` and `ping` magic strings

Tests/BotTests/Canned Data/GitHubPullRequestEvent.swift
```diff
@@ -0,0 +1,462 @@
+import Foundation
+
+let GitHubPullRequestEvent: String = """
+{
```
AliSoftware, line 4: Some indentation of this JSON fixture might be nice to help us mere humans read it too 😛

Tests/BotTests/GitHub/GitHubEventsTests.swift
```diff
@@ -0,0 +1,88 @@
...
+            )
+        )
+
+        let request = StubbedRequest(headers: ["X-GitHub-Event" : "pull_request"], body: pullRequestEvent)
+
+        var observedEvent: Event? = nil
+
+        service.events
+            .observe(on: scheduler)
+            .observeValues { event in observedEvent = event }
```
AliSoftware: A more reliable test would be to add events in an array (then assert that there's only one and that it has the expected value). That way we'd also ensure that the event is only sent once

Tests/BotTests/GitHub/GitHubEventsTests.swift
```diff
@@ -0,0 +1,88 @@
...
+        let service = GitHubService()
+        let scheduler = TestScheduler()
+
+        let request = StubbedRequest(headers: ["X-GitHub-Event" : "ping"], body: nil)
+
+        var observedEvent: Event? = nil
+
+        service.events
+            .observe(on: scheduler)
+            .observeValues { event in observedEvent = event }
```
AliSoftware: ditto

Tests/BotTests/GitHub/GitHubEventsTests.swift
```diff
@@ -0,0 +1,88 @@
...
+
+        let result = service.handleEvent(from: request)
+
+        expect(result.error) == .unknown
+    }
+}
+
+extension GitHubEventsTests {
+    fileprivate struct StubbedRequest: RequestProtocol {
+        let headers: [String : String]
```
AliSoftware: [nit]
```suggestion
        let headers: [String: String]
```

Package.swift
ilyapuchka: I like how it looks but won't this make it harder to use as a dependency?

Tests/LinuxMain.swift
```diff
@@ -0,0 +1,8 @@
+import XCTest
+
+import BotTests
+
+var tests = [XCTestCaseEntry]()
+tests += BotTests.__allTests()
```
AliSoftware, line 6: I'm guessing this was auto-generated by `swift test --generate-linuxmain`? (otherwise, not sure why we use a `var` then `+=` instead of affecting `let tests = BotTests.__allTests()` direclty... and it's sad that they don't insert a comment telling that the file was auto-generated... but I guess you can't do much about it)

Sources/App/Extensions/Vapor.Request+RequestProtocol.swift
ilyapuchka: should probably use the `name` parameter?

Sources/App/Extensions/Vapor.Request+RequestProtocol.swift
```diff
@@ -0,0 +1,14 @@
...
+    }
+
+    public func decodeBody<T>(_ type: T.Type) -> T? where T: Decodable {
+        return try? content.syncDecode(type)
```
ilyapuchka: this shouldn't be used - in Vapor 3.3.0 it results in a runtime error (not crash). should use `content.decode` and chain futures 🤷‍♂ everything is the Future when dealing with Vapor

Sources/Bot/Network/Request.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+
+public protocol RequestProtocol {
```
ilyapuchka, line 5: I'd just use an extension on `Request`, not likely it will be anything else. And I'm guessing it should be possible to create a request for tests. If not - why not to go with a struct rather than protocol?

ilyapuchka (commented):
Tbh not sure how it will play out to mix Futures which are everywhere in Vapor now and Signals. Considering that this bot only performs a single task at a time as a response to the incoming event and dies rather than is running continuously I'd maybe just go with futures...

Package.swift
dmcrodrigues: Not sure if this will be used as a dependency to be honest, this can always be changed if necessary or I can rollback this change completely.

Package.resolved
dmcrodrigues: Not necessary but the surface of testing is way smaller so performance shouldn't be a problem now and in the foreseeable future and it does improve a bit readability. I can remove it if we feel strong about this.

Sources/App/Extensions/Vapor.Request+RequestProtocol.swift
dmcrodrigues: Great catch Olivier 👍

Sources/Bot/Services/GitHubService.swift
dmcrodrigues: This can be changed since there's not really any output but I think `Result` overall provides a stronger model in terms of expression and extension, if tomorrow we add some output we could extended it without refactoring each call site.

Sources/Bot/Services/GitHubService.swift
dmcrodrigues: 👍

Tests/BotTests/Canned Data/GitHubPullRequestEvent.swift
dmcrodrigues: Poor Xcode, I have to done it externally in Visual Studio Code lol

Tests/BotTests/GitHub/GitHubEventsTests.swift
dmcrodrigues: 👍

Tests/LinuxMain.swift
dmcrodrigues: It was automatically generated, I haven't edited this file.

Sources/App/Extensions/Vapor.Request+RequestProtocol.swift
dmcrodrigues: I'd like to avoid dealing with them for now if possible.

Sources/Bot/Network/Request.swift
dmcrodrigues: The main purpose of this is to avoid having `Bot` module depending of `Vapor` at all, that would gives enough flexibility to change it to something else if needed.

Tests/LinuxMain.swift
dmcrodrigues: Interesting, `Tests/BotTests/XCTestManifests.swift` includes a comment with Xcode 10.2.1 but nothing has changed in `Tests/LinuxMain.swift`.
image

dmcrodrigues:
> Tbh not sure how it will play out to mix Futures which are everywhere in Vapor now and Signals

@ilyapuchka to avoid a huge refactor I want to keep ReactiveSwift and ensure the `Bot` does not depend on anything from Vapor or any other middleware. I think there's time to re-think this if we want/need later on.

Tests/BotTests/Canned Data/GitHubPullRequestEvent.swift
AliSoftware: Iirc Cmd-Alt-V (or Cmd-Alt-Shift-V?) pastes without reformatting so should keep the indentation

Package.swift
AliSoftware: I agree with @ilyapuchka even if it looks nice it's not ideal for a package name to have non-ascii characters.

We'll make the bot name in Slack have the nice name but as for the package I think you should revert to `WallE`. Or use `Wall-E`.

Sources/App/Extensions/Vapor.Request+RequestProtocol.swift
```diff
@@ -0,0 +1,14 @@
...
+import Bot
+import Vapor
+
+extension Vapor.Request: RequestProtocol {
+    public func header(named name: String) -> String? {
+        return http.headers[name].first
+    }
+
+    public func decodeBody<T>(_ type: T.Type) -> T? where T: Decodable {
+        return try? content.decode(type).wait()
```
AliSoftware: I have to admit I'm hesitant about the whole "Avoid using Futures just to make the code independent of Vapor" thing. Might be premature overarchitecturing which introduce main thread blocking like that just for it — even if this shouldn't be such a big thread blocker, still not good in principle.

Sources/App/Extensions/Vapor.Request+RequestProtocol.swift
ilyapuchka: The issue is that it's not safe in SwiftNIO environment, it can potentially deadlock (I don't know what is the real scenario for this to happen though). We have everything based on futures in Stevenson and it's fine, even though technically most of the performed tasks don't have to be asynchronous.

Sources/Bot/Network/Request.swift
ilyapuchka: Interesting. I was thinking about this dependency for Stevenson as well, because it does not really _have_ to use Vapor, but decided that it's fine as the app is built with Vapor anyway and the stuff that Vapor already provides even just Futures makes it not to depend on anything else, like RAS. And considering Combine I wouldn't be surprised if in future Vapor encomorates it somehow as well on top of SwiftNIO futures (which feel a bit of clumsy to use sometimes because of event loops architecture).
Let's see how it goes 👍

Sources/App/Extensions/Vapor.Request+RequestProtocol.swift
ilyapuchka: the thing is that it still does the same what asyncDecode does where the problem is in `wait` called on the event loop (iirc SwiftNIO docs explicitly say that there should be no blocking calls when code is running in the event loop) 🤷‍♂ I was baffled by that myself and have a branch in Stevenson where actually replaced use of `asyncDecode` with just `decode` (we have futures everywhere already so it fits in well)
Probably to avoid that this method should return a signal or signal producer if you want to avoid depending on Futures.
See the SwiftNIO code for reference (it's part of Vapor so you can just search for it in Xcode) - `EventLoopFuture.swift:799`.

Package.swift
dmcrodrigues: image

Sources/App/Extensions/Vapor.Request+RequestProtocol.swift
dmcrodrigues: We are now using `SignalProducer`s which start on their own scheduler, please let me know if you see any potential issue or theres'a a better way to execute a future.

Package.swift
AliSoftware: Sorry.
Sad Wall·E

dmcrodrigues: It's fine 😄 Wall-E doesn't have feelings yet so 😂

ilyapuchka (approved):

dmcrodrigues:
@AliSoftware can we merge this one?

Sources/Bot/Network/Request.swift
AliSoftware: yeah tbh trying not to depend on Vapor but then depend on RAS instead kinda defeat the initial idea, doesn't it?

Right now your goal is if we decide to use something other than Vapor we can easily switch... but if we decide to use something other than RAS, e.g. Combine, we'd still need to switch too... and having to balance with both the `Future` and `SignalProducer` concepts back and forth makes the codebase more confusing imho.
Also my fear is that there's also a risk of adding bugs if there is some convoluted race condition between Vapor's EventLoop and RAS' scheduler that we didn't think about...

Personally I don't think bringing a whole dependency just for that (Is there any other place where we use RAS instead of Futures in the Bot?), and adding a separation layer here that we won't likely need anyway (YAGNI) feels unnecessary, harder to debug and to reason about...

dmcrodrigues: @AliSoftware I agree with you but there's a limitation here, the current implementation of the bot uses RAS deeply and I'd like to have it here first up and running before making any significant refactor, that's why I'm bringing RAS but it's very likely that it will go away in the short-term. Not depending of Vapor was super easy to achieve that's why it's being done already, RAS will be done later on if it makes sense.

AliSoftware: Ok. In that case let's ship it that way then.

Still confused by your "the current implementation of the bot uses RAS deeply" since `ReactiveSwift` was added in the `Package.swift` in this very PR, so how could the implementation _before that PR_ depend on RAS then? (Or did you mean "just the current implementation of GitHubService, that I just added in this PR"?)

dmcrodrigues: Sorry, my plan is to move our current implementation (which is private ATM) here and start working from there instead of starting from scratch.

AliSoftware (approved):

PR #8, opened 2019-06-05
Verify the signature of any incoming payload to certify its origin

dmcrodrigues:
### Why?

Every payload coming from webhooks should be signed at the origin and verified when received to certify its origin, as a security procedure. This is done by verifying the body of the request with the secret configured by us on GitHub.

The reference for this verification can be found [here](https://developer.github.com/webhooks/securing/).

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -29,6 +29,15 @@
         }
       },
       {
+        "package": "CryptoSwift",
+        "repositoryURL": "https://github.com/krzyzanowskim/CryptoSwift.git",
+        "state": {
+          "branch": null,
+          "revision": "3a2acbb32ab68215ee1596ee6004da8e90c3721b",
+          "version": "1.0.0"
+        }
+      },
+      {
         "package": "DatabaseKit",
         "repositoryURL": "https://github.com/vapor/database-kit.git",
         "state": {
--- a/Sources/App/configure.swift
+++ b/Sources/App/configure.swift
@@ -1,10 +1,17 @@
 import Bot
 import Vapor
 
+enum ConfigurationError: Error {
+    case missingCOnfiguration(message: String)
+}
+
 /// Called before your application initializes.
 public func configure(_ config: inout Config, _ env: inout Environment, _ services: inout Services) throws {
 
-    let gitHubService = GitHubService()
+    guard let githubWebhookSecret = Environment.githubWebhookSecret
+        else { throw ConfigurationError.missingCOnfiguration(message: "💥 GitHub Webhook Secret is missing")}
+
+    let gitHubService = GitHubService(signatureToken: githubWebhookSecret)
 
     // Register routes to the router
     let router = EngineRouter.default()
--- a/Sources/App/configure.swift
+++ b/Sources/App/configure.swift
@@ -2,14 +2,14 @@ import Bot
 import Vapor
 
 enum ConfigurationError: Error {
-    case missingCOnfiguration(message: String)
+    case missingConfiguration(message: String)
 }
 
 /// Called before your application initializes.
 public func configure(_ config: inout Config, _ env: inout Environment, _ services: inout Services) throws {
 
     guard let githubWebhookSecret = Environment.githubWebhookSecret
-        else { throw ConfigurationError.missingCOnfiguration(message: "💥 GitHub Webhook Secret is missing")}
+        else { throw ConfigurationError.missingConfiguration(message: "💥 GitHub Webhook Secret is missing")}
 
     let gitHubService = GitHubService(signatureToken: githubWebhookSecret)
 
--- a/Sources/Bot/Services/GitHubService.swift
+++ b/Sources/Bot/Services/GitHubService.swift
@@ -55,6 +55,8 @@ public final class GitHubService {
         with token: Token
     ) -> (RequestProtocol) -> Result<RequestProtocol, EventHandlingError> {
 
+        // This was implemented following this reference: https://developer.github.com/webhooks/securing/
+
         return { request in
             guard
                 let signature = request.header(.signature),
--- a/Sources/Bot/Services/GitHubService.swift
+++ b/Sources/Bot/Services/GitHubService.swift
@@ -17,12 +17,12 @@ public final class GitHubService {
 
     public typealias Token = String
 
-    private let scheduler: QueueScheduler
+    private let scheduler: Scheduler
     private let signatureVerifier: (RequestProtocol) -> Result<RequestProtocol, EventHandlingError>
     private let eventsObserver: Signal<Event, NoError>.Observer
     public let events: Signal<Event, NoError>
 
-    public init(signatureToken: Token, scheduler: QueueScheduler = QueueScheduler()) {
+    public init(signatureToken: Token, scheduler: Scheduler = QueueScheduler()) {
         self.signatureVerifier = GitHubService.signatureVerifier()(signatureToken)
         self.scheduler = scheduler
         (events, eventsObserver) = Signal.pipe()
--- a/Sources/Bot/Services/GitHubService.swift
+++ b/Sources/Bot/Services/GitHubService.swift
@@ -23,7 +23,7 @@ public final class GitHubService {
     public let events: Signal<Event, NoError>
 
     public init(signatureToken: Token, scheduler: Scheduler = QueueScheduler()) {
-        self.signatureVerifier = GitHubService.signatureVerifier()(signatureToken)
+        self.signatureVerifier = GitHubService.signatureVerifier(with: signatureToken)
         self.scheduler = scheduler
         (events, eventsObserver) = Signal.pipe()
     }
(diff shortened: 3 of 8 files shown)


Tests/BotTests/GitHub/GitHubEventsTests.swift
```diff
@@ -2,33 +2,47 @@ import XCTest
...
+
+    private func signature(for data: Data) -> String {
+        let signature = try! HMAC(key: signatureToken, variant: .sha1)
+            .authenticate(data.bytes)
+            .toHexString()
+
+        return "sha1=\(signature)"
+    }
+
     func test_handling_pull_request_event() {
```
dmcrodrigues: I'll refactor the tests to make them more idiomatic.

Package.resolved
```diff
@@ -28,6 +28,15 @@
           "version": "3.3.2"
         }
       },
+      {
+        "package": "CryptoSwift",
```
dmcrodrigues, line 32: NOTE: I've included `CryptoSwift` to avoid depending on Vapor directly for the same reasons expressed in #7.

Sources/App/configure.swift
```diff
@@ -1,10 +1,17 @@
 import Bot
 import Vapor
 
+enum ConfigurationError: Error {
+    case missingCOnfiguration(message: String)
```
AliSoftware:
```suggestion
    case missingConfiguration(message: String)
```

Sources/App/configure.swift
```diff
@@ -1,10 +1,17 @@
...
+}
+
 /// Called before your application initializes.
 public func configure(_ config: inout Config, _ env: inout Environment, _ services: inout Services) throws {
 
-    let gitHubService = GitHubService()
+    guard let githubWebhookSecret = Environment.githubWebhookSecret
+        else { throw ConfigurationError.missingCOnfiguration(message: "💥 GitHub Webhook Secret is missing")}
```
AliSoftware:
```suggestion
        else { throw ConfigurationError.missingConfiguration(message: "💥 GitHub Webhook Secret is missing")}
```

Sources/Bot/Services/GitHubService.swift
```diff
@@ -43,10 +50,37 @@ public final class GitHubService {
             return SignalProducer(error: .unknown)
         }
     }
+
+    private static func signatureVerifier(
+    ) -> (Token) -> (RequestProtocol) -> Result<RequestProtocol, EventHandlingError> {
```
AliSoftware: Why 3 levels of currying instead of two? Wouldn't this be enough?

```swift
private static func signatureVerifier(token: Token) -> (RequestProtocol) -> Result<RequestProtocol, EventHandlingError>
```

_And then in the `init` you could avoid the awkward `()(signatureToken)` syntax and just do this:_
```swift
self.signatureVerifier = GitHubService.signatureVerifier(signatureToken)
```

Sources/Bot/Services/GitHubService.swift
```diff
@@ -43,10 +50,37 @@ public final class GitHubService {
...
+
+        return { token in
+            return { request in
```
AliSoftware: Would be nice to put a comment pointing to the GitHub algorithm description (i.e. [the link you put in the PR body](https://developer.github.com/webhooks/securing/)) somewhere there

Sources/Bot/Services/GitHubService.swift
```diff
@@ -43,10 +50,37 @@ public final class GitHubService {
...
+                guard
+                    let signature = request.header(.signature),
+                    let digest = signature.range(of: "sha1=")
+                        .map({ signature[$0.upperBound..<signature.endIndex] })
+                        .map(String.init)
+                        .map(Array.init(hex:))
```
AliSoftware: 🤔❔ Where does `Array(hex:)` comes from, is that provided by `CryptoSwift`?

Tests/BotTests/GitHub/GitHubEventsTests.swift
```diff
@@ -2,33 +2,47 @@ import XCTest
...
                 mergeState: .clean
             )
         )
 
         let request = StubbedRequest(
-            headers: [GitHubService.HTTPHeader.event.rawValue: GitHubService.APIEvent.pullRequest.rawValue],
-            body: pullRequestEvent
+            headers: [
+                GitHubService.HTTPHeader.event.rawValue: GitHubService.APIEvent.pullRequest.rawValue,
+                GitHubService.HTTPHeader.signature.rawValue: signature(for: GitHubPullRequestEvent.data(using: .utf8)!)
```
AliSoftware: Thinking out loud: doesn't it defeat the purpose to call the method computing the `signature(for:)` dynamically as a fixture for the test? I mean, it feels like checking if `func f(_ x: Int) { x * 2 }` works by testing `expect(f(2)) == f(2)` instead of testing `expect(f(2)) == 4`, no?

Maybe it's more resilient to put the hardcoded expected signature there?

Tests/BotTests/GitHub/GitHubEventsTests.swift
```diff
@@ -69,27 +89,49 @@ class GitHubEventsTests: XCTestCase {
...
+
+        let request = StubbedRequest(
+            headers: [GitHubService.HTTPHeader.event.rawValue : "anything_really"],
+            data: nil
+        )
+
+        let result = service.handleEvent(from: request).first()
+
+        expect(result?.error) == .untrustworthy
+    }
```
AliSoftware, line 157: This one mainly test if a request _doesn't have_ the expected signature header at all, but we should also have a test with the header present but an invalid value (no `sha=` prefix) and another one with an invalid signature (`sha=someWellFormedSHAButNotTheExpectedSignature`)

Tests/BotTests/GitHub/GitHubEventsTests.swift
```diff
@@ -99,9 +141,14 @@ extension GitHubEventsTests {
...
-            case let .some(value):
-                return SignalProducer(value: value)
+            switch data {
+            case let .some(data):
+                let decoder = JSONDecoder()
+                do {
+                    return SignalProducer(value: try decoder.decode(T.self, from: data))
+                } catch {
+                    return SignalProducer(error: AnyError(DecodeError.invalid))
+                }
```
AliSoftware: [nit] just a style question, why the `switch` and not a `guard let data = data else { return SignalProducer(error: AnyError(DecodeError.invalid)) }` early return?

AliSoftware (commented):
❤️ Great addition 👍

Sources/Bot/Services/GitHubService.swift
dmcrodrigues: This makes no sense, you're right.

Sources/Bot/Services/GitHubService.swift
dmcrodrigues: Yeah, CryptoSwift's sugar.

Tests/BotTests/GitHub/GitHubEventsTests.swift
dmcrodrigues: I think you have a point here Olivier 👍

Tests/BotTests/GitHub/GitHubEventsTests.swift
dmcrodrigues: No reason really, I've changed it Olivier thanks.

Tests/BotTests/GitHub/GitHubEventsTests.swift
dmcrodrigues: You're right, better be safe 👍

Tests/BotTests/GitHub/GitHubEventsTests.swift
```diff
@@ -2,83 +2,170 @@ import XCTest
...
-        let scheduler = TestScheduler()
+extension GitHubEventsTests {
+    
+    static let signatureToken = "Wall-E"
+
+    private func signature(for data: Data) -> String {
+        return try! HMAC(key: GitHubEventsTests.signatureToken, variant: .sha1)
+            .authenticate(data.bytes)
+            .toHexString()
+    }
```
AliSoftware: Is this private method still used now that we use hardcoded signatures in fixtures?

Sources/Bot/Services/GitHubService.swift
```diff
@@ -1,29 +1,36 @@
...
     internal enum APIEvent: String {
         case pullRequest = "pull_request"
         case ping
     }
 
-    private let scheduler: QueueScheduler
+    public typealias Token = String
+
+    private let scheduler: Scheduler
+    private let signatureVerifier: (RequestProtocol) -> Result<RequestProtocol, EventHandlingError>
```
AliSoftware, line 21: Is this still required to have that injected as a closure now that we don't inject a signatureVerifier in tests?

Might as well remove the closure property here and use the private method directly

AliSoftware (commented):
_(reviewed from 📱might have oversaw things)_

Tests/BotTests/GitHub/GitHubEventsTests.swift
dmcrodrigues: It's useful if we need to update/add new signatures that's why I kept it, I can leave a comment.

Sources/Bot/Services/GitHubService.swift
dmcrodrigues: I don't want to keep the token accessible in memory, capturing it in the closure makes it harder to be accessed although as everything not impossible.

Tests/BotTests/GitHub/GitHubEventsTests.swift
AliSoftware: I see. Well to be fair we could just do like we do when re recording snapshot tests, but manually: run them, see them fail, and copy/paste the "actual" value obtained during test run in place of the "expected" value… no need for the private method for that 😅

dmcrodrigues: 👍 I will remove it then.

Sources/Bot/Services/GitHubService.swift
AliSoftware: Ah, good point 👍

Tests/BotTests/GitHub/GitHubEventsTests.swift
```diff
@@ -2,83 +2,164 @@ import XCTest
...
+                    with: .ping,
+                    signature: "9f64868b5cd91faa4c63589acff7286b16d289ae",
+                    payload: "{}"
+                )
+
+                let result = service.handleEvent(from: req).first()
+
+                scheduler.advance()
+
+                expect(result?.error).to(beNil())
```
AliSoftware: Shouldn't all those `expect` be grouped in the assert closure (which you'd tweak to also pass the result as parameter)?

Tests/BotTests/GitHub/GitHubEventsTests.swift
```diff
@@ -2,83 +2,164 @@ import XCTest
...
+        let service = stub(scheduler)
 
         var observedEvents: [Event] = []
 
         service.events
             .observe(on: scheduler)
             .observeValues { event in observedEvents.append(event) }
 
-        let result = service.handleEvent(from: request).first()
+        when(service, scheduler)
```
AliSoftware: Maybe make `let result = when(service, scheduler)` then pass the result to `assert` too, so that the assertions on `result.error` can be moved to the assert closure. Would fit better with the Arrange/Act/Assert pattern

AliSoftware (commented):
Looking good!

Sorry another round of feedback though, something that I should have spotted before, my bad. No big deal though, just a nit change; should be good to go after that for me.

Tests/BotTests/GitHub/GitHubEventsTests.swift
dmcrodrigues: Probably makes sense, let's do it.

AliSoftware (approved):

AliSoftware:
Great job @dmcrodrigues 👏

ilyapuchka:
I'm pretty sure @AliSoftware 's review is more than sufficient here, so I'll leave this PR up to you guys 👍

ilyapuchka (approved):

PR #9, opened 2019-06-06
Logging

dmcrodrigues:
### Why?

Include a very simple logger just to keep track of any activity, e.g. incoming requests.


ilyapuchka (approved):

AliSoftware (approved):
I wondered if we couldn't have just used a `logger: (String) -> ()` closure instead of a `protocol` for simplicity. But in the end I prefer the `protocol` like you did, as it's more explicit... and can evolve to support more methods (e.g. `warn`/`error`) later, so 👍

PR #12, opened 2019-07-19
Add `MergeService` to make the bot run

dmcrodrigues:
Ticket: https://babylonpartners.atlassian.net/browse/CNSMR-2254

## Why?

We are in the process of moving our existent bot to be this repository to make it OSS and continue development with further improvements and new features.

## How?

This keeps the same existent infrastructure and brings the existent models except for some namespaces what were removed. `PullRequestContext` was renamed to `PullRequestMetadata` as this model was already added here. Other than that everything was kept the same and only the following PRs will start to introduce changes.

## What's missing?

~I still need to implement the API layer to integrate with GitHub and then pipe the events to the service to make it work. I'm hoping to finish this very soon.~

Due the significant number of changes that this will require, it's better to have them in a separate PR.

PR #13, opened 2019-07-22
Adopt the usage of services

dmcrodrigues:
## Why?

Disclaimer: this change is needed for the work being done in #12 and only affects the current implementation/infrastructure so can and should be reviewed in insolation. 

`GitHubService` will require a HTTP client to interact with the GitHub API as part of the work being done in #12 which is included in Vapor so we need to make a couple of changes to accommodate its usage.

## How?

The `HTTPClient` part of Vapor needs to, or ideally should, be created from a `Container` which can be either the `Application` or the `Request`. Because `GitHubService` will be an instance that we want to set-up once and reuse in any incoming request the only good candidate for our use case would be using the `Application`. But we can't because `Application` is only created after we configure and register everything and we can't register a new additional service.

The only alternative is to use Vapor's [Services](https://docs.vapor.codes/3.0/getting-started/services/#environment) which are in essence a `ServiceLocator` providing methods to register instances or factory methods that receive the inner `Container` allowing the correct initialisation and configuration of each Service.

Adopting this in our case is relatively trivial, instead of carrying the dependencies manually to the routing system we can register each service and then retrieve them for each request as needed. Because in #12 we will need the container to set-up a HTTP client we are already using the factory method that receives the container and creates the service.

`GitHubService` in this case will be a lazy service, it will be initialized on first request and then reused in the following ones which fits our purpose of keeping just a single entity during the entire lifecycle of the app.

### DIFF
--- a/Sources/App/app.swift
+++ b/Sources/App/app.swift
@@ -10,7 +10,3 @@ public func app(_ env: Environment) throws -> Application {
     try boot(app)
     return app
 }
-
-extension Environment {
-    static let githubWebhookSecret = Environment.get("GITHUB_WEBHOOK_SECRET")
-}
--- a/Sources/App/configure.swift
+++ b/Sources/App/configure.swift
@@ -12,14 +12,21 @@ public func configure(_ config: inout Config, _ env: inout Environment, _ servic
 
     logger.log("👟 Starting up...")
 
-    guard let githubWebhookSecret = Environment.githubWebhookSecret
-        else { throw ConfigurationError.missingConfiguration(message: "💥 GitHub Webhook Secret is missing")}
+    services.register(logger, as: LoggerProtocol.self)
 
-    let gitHubService = GitHubService(signatureToken: githubWebhookSecret)
+    services.register(GitHubService.self) { container -> GitHubService in
+
+        logger.log("⏳ Initializing `GitHubService`...")
+
+        guard let githubWebhookSecret = Environment.githubWebhookSecret
+            else { throw ConfigurationError.missingConfiguration(message: "💥 GitHub Webhook Secret is missing")}
+
+        return GitHubService(signatureToken: githubWebhookSecret)
+    }
 
     // Register routes to the router
     let router = EngineRouter.default()
-    try routes(router, logger: logger, gitHubService: gitHubService)
+    try routes(router)
     services.register(router, as: Router.self)
 
     // Register middleware
@@ -30,3 +37,10 @@ public func configure(_ config: inout Config, _ env: inout Environment, _ servic
 
     logger.log("🏁 Ready")
 }
+
+extension Environment {
+    static let githubWebhookSecret = Environment.get("GITHUB_WEBHOOK_SECRET")
+}
+
+extension GitHubService: Service {}
+extension PrintLogger: Service {}
--- a/Sources/App/routes.swift
+++ b/Sources/App/routes.swift
@@ -1,13 +1,16 @@
 import Bot
 import Vapor
 
-public func routes(_ router: Router, logger: LoggerProtocol, gitHubService: GitHubService) throws {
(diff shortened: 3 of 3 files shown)


dmcrodrigues:
We don't have PullAssigners configured yet so I requested reviews randomly 🙈 (except for @ilyapuchka)

Sources/App/app.swift
```diff
@@ -10,7 +10,3 @@ public func app(_ env: Environment) throws -> Application {
     try boot(app)
     return app
 }
-
-extension Environment {
```
dmcrodrigues, line 14: This was moved to be near the place where it's used, nothing else.

RuiAAPeres (approved):

Sources/App/configure.swift
```diff
@@ -12,14 +12,21 @@ public func configure(_ config: inout Config, _ env: inout Environment, _ servic
...
-        else { throw ConfigurationError.missingConfiguration(message: "💥 GitHub Webhook Secret is missing")}
+    services.register(logger, as: LoggerProtocol.self)
 
-    let gitHubService = GitHubService(signatureToken: githubWebhookSecret)
+    services.register(GitHubService.self) { container -> GitHubService in
+
+        logger.log("⏳ Initializing `GitHubService`...")
+
+        guard let githubWebhookSecret = Environment.githubWebhookSecret
+            else { throw ConfigurationError.missingConfiguration(message: "💥 GitHub Webhook Secret is missing")}
```
ilyapuchka, line 22:
```suggestion
            else { throw ConfigurationError.missingConfiguration(message: "💥 GitHub Webhook Secret is missing") }
```

Btw we have some convenience extensions for environment variables in Stevenson as we have to use quite a few
I wonder if we should at some point consider using Stevenson as a dependency on Wall-E for GitHub service as it also needs it, but keep the rest of logic in Wall-E itself, like we do with Stevenson app target.

ilyapuchka (approved):

ilyapuchka:
In Stevenson rather than using Services we went with passing these service dependencies directly in the routes function as at this point they are already available and capturing them in routes. I guess it achieves the same result but won't query DI container on every call to route.

ilyapuchka:
And speaking of HTTP client we always pass a container to the functions that need i.e. to get HTTP client (which might not be the best and should be probably moved out of these functions to allow testability)

dmcrodrigues:
> In Stevenson rather than using Services we went with passing these service dependencies directly in the routes function as at this point they are already available and capturing them in routes. I guess it achieves the same result but won't query DI container on every call to route.

This is what we were doing but `GitHubService`, which should be alive for the entire lifetime of the app, will require a HTTP client very soon and at the time it was being created we don't have either an application or request so the only way to ensure we can pass the container seems to be using services. Do you see a better way to achieve this?

Sources/App/configure.swift
dmcrodrigues: I think we should analyse that option, it makes sense to reuse some of the infrastructure (at least) between them but I think we should do it after getting it running and addressing the issues that we would like to get fixed as soon as possible.

ilyapuchka:
@dmcrodrigues I don't think one way is better than another, but for example in CircleCIService we pass the request object (it is a container) from the route to the method in the service that is then getting HTTP client from it with `container.client()` (that's what I mean should probably change, though in test we can just register fake client) - https://github.com/Babylonpartners/Stevenson/blob/master/Sources/Stevenson/CircleCIService.swift#L41
If you have a case when you need to access the client before the incoming request happens then I guess there is no other option indeed

dmcrodrigues:
I'm blocking this because this changes may not be necessary after all.

mr-v:
> I'm blocking this because this changes may not be necessary after all.

Can this be closed?

dmcrodrigues:
We no longer need this changes so closing this one.

PR #14, opened 2019-07-25
Complete the transition to make Wall-E ready to be used

dmcrodrigues:
Ticket: https://babylonpartners.atlassian.net/browse/CNSMR-2254

## Why?

This is a follow-up of #12 and includes the integration with the GitHub API necessary to complete the transition and make the bot ready to be used.

## How?

The bot is being developed in a separate framework and because of that we are using `URLSession` from `Foundation` over using the http client from Vapor. The main purpose of this separation is to ensure that we can embed the bot in any infrastructure, existent or new, if needed.

The GitHub API is implemented based on the concept of resources which can be requested to the `GitHubClient`. These resources can be obtained from factories like `Repository` to ensure the path is correctly defined.

There was a couple of changes in `GitHubAPIProtocol` which partially necessary to accommodate either the new models defined in Wall-E or to support the decoding. There's a couple of changes and improvements to be done in this interface but that should be done in a separate PR to avoid prolonging the readiness of the bot. 

Testing is being done with Vinyl through fixtures.

## What's missing?

There's a couple of methods to implement but the overall infrastructure is already settled for review.


Sources/Bot/API/RepositoryAPI.swift
```diff
@@ -0,0 +1,111 @@
...
+            .mapError(AnyError.init)
+    }
+
+    public func fetchCommitStatus(for pullRequest: PullRequest) -> SignalProducer<CommitState, AnyError> {
+        return client.request(repository.commitStatus(for: pullRequest))
+            .mapError(AnyError.init)
+    }
+
+    public func merge(intoBranch branch: String, head: String) -> SignalProducer<MergeResult, AnyError> {
+        fatalError()
```
AliSoftware: Add a TODO: with ticket number or a message in fatalError, in order to be sure not to forget to implement that later

Sources/Bot/API/Response.swift
```diff
@@ -0,0 +1,66 @@
+
+// Copied from https://github.com/mdiep/Tentacle
+
+import Foundation
+
+let LinksRegex = try! NSRegularExpression(pattern: "(?<=\\A|,) *<([^>]+)>( *; *\\w+ *= *\"[^\"]+\")* *(?=\\z|,)")
+let LinkParamRegex = try! NSRegularExpression(pattern: "; *(\\w+) *= *\"([^\"]+)\"")
```
AliSoftware: lowercase (even for constants), to match Apple guidelines.

Sources/Bot/API/Response.swift
```diff
@@ -0,0 +1,66 @@
...
+
+            relName = String(params[Range(match.range(at: 2), in: params)!])
+        }
+
+        if let relName = relName {
+            links[relName] = url
+        }
+    }
+    return links
+}
```
AliSoftware, line 30: This might deserve a unit test

Tests/BotTests/MergeServiceTests.swift
```diff
@@ -10,14 +10,14 @@ struct MockLogger: LoggerProtocol {
     }
 }
 
-let IntegrationLabel = "Please Merge 🙏"
+let IntegrationLabel = PullRequest.Label(name: "Please Merge 🙏")
```
AliSoftware: lowercase

AliSoftware (commented):
Preliminary review (I know, PR still WIP 😉 )

Sources/Bot/API/RepositoryAPI.swift
dmcrodrigues: This will be implemented as part of this PR @AliSoftware, I'm working on it.

Sources/Bot/API/Response.swift
dmcrodrigues: I brought this from [Tentacle](https://github.com/mdiep/Tentacle) so I was trying to keep it unmodified but I'm actually changing this to remove the stuff we don't need and structure to fit our purpose.

Sources/Bot/API/Response.swift
dmcrodrigues: 👍

Tests/BotTests/MergeServiceTests.swift
dmcrodrigues: 👍

Sources/Bot/API/Response.swift
AliSoftware: Just curious: why don't we depend on Tentacle directly instead is reinventing the wheel? NIH?

dmcrodrigues: I tried @AliSoftware but our models are slight different, mainly `PullRequest` which only contains the url links for the pull request, diff, ... 

Also another problem I found out was with extending it, `Resource` does not have a public initialiser so we can't extend to other APIs that we need like fetching all pull requests from a repository.

Based on the structure we are using I think we can migrate to it if we address this limitations but I didn't want to wait more time on this so it was faster to develop our own infra-structure.

AliSoftware: Ok. I think it's worth opening an issue on the repo at least to ask for those changes and making the init public. Will ping the author in Slack too 😉

dmcrodrigues: 👍 I was planning to open a ticket anyway but I'd like to get this one done first 😅

mdiep: @dmcrodrigues I'm happy to add you as a contributor to Tentacle too if you're interested in making improvements. :)

dmcrodrigues: @mdiep I'm definitely interested ❤️  I didn't take that route from the beginning because I was short on time but I based most of our implementation to ensure we could switch in the future.

Sources/Bot/API/Resource.swift
```diff
@@ -0,0 +1,41 @@
...
+
+internal struct Resource<T> {
+    typealias Decoder = (Response) -> Result<T, GitHubClient.Error>
+
+    internal var method: Method
+    internal var path: String
+    internal var queryItems: [URLQueryItem]
+    internal var headers: [String: String]
+    internal var body: Data?
+    internal let decoder: Decoder
```
dmcrodrigues, line 12: This was necessary to support the merge of branches which can finish with three valid codes:

- 200: Merged
- 204: Nothing to merge
- 409: Conflict

I tried to merge `200` and `204` in a resource that would return an optional `Commit` but that doesn't work because with a `204` we don't receive a body leading to a failure during decoding.

Sources/Bot/API/Response.swift
dmcrodrigues: We have two tests now to certify we parse this correctly.

Sources/Bot/API/GitHubAPIProtocol.swift
```diff
@@ -1,21 +1,22 @@
 import ReactiveSwift
 import Result
 
+// TODO: Refactor interface
```
dmcrodrigues, line 4: I tried to minimise the changes here but some where helpful to avoid unnecessary mappings.

dmcrodrigues:
Ok this one is now ready to be reviewed, I'm not going to include the actual integration of the `MergeService` to make it run in this PR to avoid making it more complex and bigger than what it already is. The main purpose of the PR is to integrate with the GitHub API.

Sources/Bot/API/GitHubAPIProtocol.swift
```diff
@@ -1,21 +1,22 @@
...
 
     func fetchPullRequest(number: UInt) -> SignalProducer<PullRequestMetadata, AnyError>
 
-    func fetchCommitStatus(for pullRequest: PullRequest) -> SignalProducer<CommitStatus, AnyError>
+    func fetchCommitStatus(for pullRequest: PullRequest) -> SignalProducer<CommitState, AnyError>
 
-    func merge(intoBranch branch: String, head: String) -> SignalProducer<MergeResult, AnyError>
+    func performMerge(base: PullRequest.Branch, head: PullRequest.Branch) -> SignalProducer<MergeResult, AnyError>
 
     func mergePullRequest(_ pullRequest: PullRequest) -> SignalProducer<(), AnyError>
```
AliSoftware: The purpose of this one is not clear to me (at least not with the name alone, if I want to know what it's expected to do I'd need to go to the impl) as opposed to `performMerge(base:head:)`. Is `mergePullRequest` just extracting the base and head branches from `PullRequest` then calling `performMerge` on it?
Maybe a doc-comment to explain the difference between the two and why we'd call one instead of the other would help.

Sources/Bot/API/GitHubClient.swift
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private var authorizationHeaders: () -> [String : String]
+
+    public init(session: URLSession = .shared, token: String) {
+        self.session = session
+        self.authorizationHeaders = { token.isEmpty ? [:] : ["Authorization": "token \(token)"] }
+    }
+
+    private func request(_ request: URLRequest) -> SignalProducer<Response, Error> {
```
AliSoftware: [nit]: move this to line 56 to keep public/internal methods first and private methods afterwards

Sources/Bot/API/RepositoryAPI.swift
```diff
@@ -0,0 +1,160 @@
...
+    // TODO: This should be moved to be done as part of the handling of resources in `GitHubClient`
+    private func encode<T: Encodable>(_ t: T) -> Data {
+        guard let data = try? JSONEncoder().encode(t)
+            else { fatalError("Unexpected failure while encoding `\(t)`") }
+
+        return data
+    }
+}
+
+public struct RepositoryAPI: GitHubAPIProtocol {
```
AliSoftware, line 5: [nit] split `struct Repository` and `struct RepositoryAPI` in separate files?

Sources/Bot/API/RepositoryAPI.swift
```diff
@@ -0,0 +1,160 @@
...
+
+    public init(client: GitHubClient, repository: Repository) {
+        self.client = client
+        self.repository = repository
+    }
+
+    public func fetchPullRequests() -> SignalProducer<[PullRequest], AnyError> {
+        return client.request(repository.pullRequests)
+            .flatten()
+            .collect()
```
AliSoftware, line 18: My RAS is a bit rusty, not sure why you need `.flatten` + `.collect` here?

I thought `client.request(repository.pullRequests)` would return `SignalProducer<[PullRequest]>` already... and then `.flatten` would turn that into `[SignalProducer<PullRequest>]` and `collect` would turn that back into a `SignalProducer<[PullRequest]>`, right? Is that a RAS way to sync all the signals (one for each pages) emitted into a single signal emitting a single value? (if so I'm surprised there's no dedicated RAS operator? But again, my RAS-fu is a bit lacking)

Sources/Bot/API/RepositoryAPI.swift
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    public func mergePullRequest(_ pullRequest: PullRequest) -> SignalProducer<(), AnyError> {
+        return client.request(repository.merge(pullRequest: pullRequest))
+            .mapError(AnyError.init)
+    }
+
+    public func deleteBranch(named branch: PullRequest.Branch) -> SignalProducer<(), AnyError> {
+        return client.request(repository.deleteBranch(branch: branch))
+            .map { _ in }
```
AliSoftware: `deleteBranch(branch:)` already returns a `Resource<Void>` so `client.request` should already return `SignalProducer<Void, Error>` shouldn't it? So why the `.map { _ in }` there?

Also, [nit] I thought the convention for return types was to use `Void` not `()`, so I'd expect all those methods to be `SignalProducer<Void, AnyError>`? (not a big deal, but it's just about conventions)

Sources/Bot/Extensions/JSONDecoderExtensions.swift
```diff
@@ -0,0 +1,25 @@
...
+    let formatter = DateFormatter()
+    formatter.locale = Locale(identifier:"en_US_POSIX")
+    formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss'Z'"
+    formatter.timeZone = TimeZone(abbreviation:"UTC")
+    return formatter
+}()
+
+extension JSONDecoder {
+
+    static var iso8601Decoder: JSONDecoder {
```
AliSoftware: [nit] the name feels super-strange to me, first time I read the call site I wondered what that meant, in the sense that at call site it looks like the JSON decoder is the one being ISO8601, like if that ISO standard was about JSON.

Maybe just renaming this to `static var withISO8601: JSONDecoder` would fit better?

Sources/Bot/Extensions/URL+QueryItems.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+extension URL {
+    func addingQueryItems(_ queryItems: [URLQueryItem]) -> URL {
+        guard queryItems.isEmpty == false else { return self }
+        
+        var components = URLComponents(url: self, resolvingAgainstBaseURL: true)!
+        components.queryItems = components.queryItems.map { $0 + queryItems } ?? queryItems
```
AliSoftware: I find this to be more readable, but in the end it's just a matter of taste, so... up to you.
```suggestion
        components.queryItems = (components.queryItems ?? []) + queryItems
```

Sources/Bot/Models/CommitState.swift
```diff
@@ -0,0 +1,19 @@
...
+
+public struct CommitState: Decodable, Equatable {
+
+    public enum State: String, Decodable {
+        case pending
+        case failure
+        case success
+    }
+
+    public struct Statuses: Equatable, Decodable {
```
AliSoftware: [nit] Feels strange that the type is named in plural while it only represents a single instance.
It's only plural because you use it in an array down below, but an array of strings is `[String]` not `[Strings]`, so an array of Statuses should be `let statuses: [Status]` not `let statuses: [Statuses]` :wink:

Tests/BotTests/GitHub/GitHubAPITests.swift
```diff
@@ -0,0 +1,286 @@
...
+                        )
+                    )]
+            )
+        ) { client in
+
+            let api = RepositoryAPI(client: client, repository: .init(owner: "golang", name: "go"))
+
+            let result: Void? = api.deleteBranch(named: target.source).first()?.value
+
+            expect(result).toNot(beNil())
```
AliSoftware, line 115: Should we also have a test checking when `deleteBranch` (and all other similar tests) returns an error? (Or are error responses handled in a more global test that I missed?)

Tests/BotTests/GitHub/ResponseTests.swift
```diff
@@ -0,0 +1,29 @@
...
+import Nimble
+@testable  import Bot
+
+class ResponseTests: XCTestCase {
+
+    func test_decoding_with_next_page_available() {
+        let response =  Response(
+            statusCode: 200,
+            headers: [
+                "Link": "<https://api.github.com/repositories/23096959/pulls?page=2&per_page=100>; rel=\"next\", <https://api.github.com/repositories/23096959/pulls?page=2&per_page=100>; rel=\"last\""
```
AliSoftware, line 12:
```suggestion
                "Link": #"<https://api.github.com/repositories/23096959/pulls?page=2&per_page=100>; rel="next", <https://api.github.com/repositories/23096959/pulls?page=2&per_page=100>; rel="last""#
```
(not sure which one I prefer tbh, up to you to accept this suggestion or not)

Tests/BotTests/GitHub/ResponseTests.swift
```diff
@@ -0,0 +1,29 @@
...
+            body: Data())
+
+        expect(response.containsReferenceForNextPage) == true
+    }
+
+    func test_decoding_without_more_pages() {
+        let response =  Response(
+            statusCode: 200,
+            headers: [
+                "Link": "<https://api.github.com/repositories/23096959/pulls?page=1&per_page=100>; rel=\"prev\", <https://api.github.com/repositories/23096959/pulls?page=1&per_page=100>; rel=\"first\""
```
AliSoftware, line 23:
```suggestion
                "Link": #"<https://api.github.com/repositories/23096959/pulls?page=1&per_page=100>; rel="prev", <https://api.github.com/repositories/23096959/pulls?page=1&per_page=100>; rel="first""#
```
(not sure which one I prefer tbh, up to you to accept this suggestion or not)

Tests/BotTests/GitHub/GitHubAPITests.swift
```diff
@@ -0,0 +1,286 @@
+import XCTest
+import Nimble
+@testable import Bot
+
+class GitHubAPITests: XCTestCase {
+    private var isRecoding = false
```
AliSoftware: Unused, to remove?

Tests/BotTests/Interceptor/Coordinator.swift
```diff
@@ -0,0 +1,95 @@
...
+        Interceptor.stopRecording()
+    }
+
+    func load(stubs newStubs: [Interceptor.Stub]) {
+        stubs.append(contentsOf: newStubs)
+    }
+
+    func loadOrRecordStubs(into url: URL) {
+
+        if FileManager.default.fileExists(atPath: url.absoluteString) {
```
AliSoftware, line 41: So does that mean that in order to re-record new stubs, there's no flag to turn on/off (like `self.recordMode = true`) in the code? It will just be in replay mode if file is present, and record mode if file is absent, so to re-record you just delete the file?

That's cool but does that also mean that if we forget to commit the recorded file, and thus the CI won't have it, it will be in record mode (thus hitting the real API) every time the CI runs, without us noticing that we forgot to commit? Or is there an automatic mechanism that makes the tests fail if we're in `recordMode`, to be sure CI is never in record mode accidentally but record mode is only used locally while working on a feature?

Tests/BotTests/Interceptor/Coordinator.swift
```diff
@@ -0,0 +1,95 @@
...
+            FileManager.default.createFile(atPath: context.destination.absoluteString, contents: data, attributes: nil)
+            else { fatalError("Failed to record stubs at `\(context.destination)`") }
+
+        recordingContext = nil
+    }
+
+    func stub(for request: URLRequest) -> Interceptor.Stub? {
+        guard stubs.isEmpty == false else { return nil }
+
+        return stubs.remove(at: 0)
```
AliSoftware, line 71: So you don't try to match the `URLRequest`? If it happens that we added the stubs for the wrong URLRequest, we won't see the issue then? (and if it's intended, and you just want to `pop` the next stub regardless of the request, why pass the `request` as parameter?)

Tests/BotTests/Interceptor/Coordinator.swift
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    func record(request: URLRequest, completion: @escaping (Interceptor.Stub) -> Void) {
+        guard let recordingContext = recordingContext
+            else { fatalError("Trying to record a stub outside of") }
```
AliSoftware:
```suggestion
            else { fatalError("Trying to record a stub outside of a recording session") }
```

Tests/BotTests/MergeServiceTests.swift
```diff
@@ -648,25 +648,25 @@ class MergeServiceTests: XCTestCase {
...
+                .postComment { message, pullRequest in
                     expect(message) == "Your pull request was accepted and it's currently `#2` in the queue, hold tight ⏳"
-                    expect(number) == 2
+                    expect(pullRequest.number) == 2
                 },
-                .postComment { message, number in
+                .postComment { message, pullRequest in
                     expect(message) == "Your pull request was accepted and it's currently `#3` in the queue, hold tight ⏳"
-                    expect(number) == 3
+                    expect(pullRequest.number) == 3
```
AliSoftware: It's disturbing that the number/position in the queue happen to be the exact same as the PR number. At first I thought "wait, `pullRequest.number` is supposed to contain the order of the PR in the queue, not the GitHub PR number?" then I realised it was just a coincidence... but if it happens that we have a regression that makes it accidentally happen (e.g. `pullRequest.number` accidentally assigned to the queue position instead of GH PR ID), we will not catch the error... so maybe just invert the order in which you merge the PRs in the stub, or assign unrelated/random PR numbers to avoid that confusion?

Tests/BotTests/Mocks/MockGitHubAPI.swift
```diff
@@ -55,7 +55,7 @@ struct MockGitHubAPI: GitHubAPIProtocol {
         }
     }
 
-    func fetchCommitStatus(for pullRequest: PullRequest) -> SignalProducer<CommitStatus, AnyError> {
+    func fetchCommitStatus(for pullRequest: PullRequest) -> SignalProducer<CommitState, AnyError> {
 
         let index = iteration.modify { iteration -> Int in
```
AliSoftware: This repeating pattern of
```
        let index = iteration.modify { iteration -> Int in
            iteration = iteration + 1
            return iteration - 1
        }
```
is hard to parse, took me a minute to understand why you'd do a `+1` then a `-1`. And I'm not sure how this `Atomic` API is supposed to be used anyway tbh (like, if the `iteration` closure parameter is `inout` so that you can do `+= 1` to modify it inline, then why does the API also expect to return a value? And if it expects to return the previous value, why didn't they automate that? I'd expect the `Atomic` API to either be `(inout Int) -> Void` or `(Int) -> Int` but not both `inout` and returning a value...

Also the name `iteration` threw me away tbh, wasn't sure what it was supposed to represent

So since it's both a bit confusing and this snippet is repeated multiple times, maybe a renaming and small private func to handle it could help. Like `stubCount` instead of `iteration`, and `func pushNextIndex() -> Int` or something to extract that snippet that both atomically increment the count and return the corresponding index?

Last but not least, not sure if `Atomic` is enough there. I mean, if 2 requests are run concurrently, you might end up incrementing your `iteration` and getting index 0 for the first, then for the second, but then the thread/runloop on which first is ran is paused while request 2 is handled, so you end up returning stub 1 before stub 0, right? At least, not totally sure if that would happen that way but my point is that it might be worth not only wrapping the index increment in an atomic operation, but even the whole access to `stubs`?

AliSoftware (commented):
Mostly nits, but also some clarification needed, some unused code to probably remove, and questions about recordMode + stubs unstacking without checking request

Sources/Bot/API/RepositoryAPI.swift
dmcrodrigues: `client.request(repository.pullRequests)` returns a `SignalProducer<[PullRequest], GitHubClient.Error>` which sends a value for each page fetched, if there's two pages we will receive two values with the contents of each page. 

As a consumer I think it's easier to receive all pull requests over multiple values representing each page, that's why we flatten everything and collect to gather everything in a single list but can be removed and then we can do it explicitly at call site.

AliSoftware: Ok so that's what I figured then, thanks a lot for the clarification. I'm surprised there's no single operator in RAS for that, but if not then `.flatten().collect()` makes sense 👍 
(bonus points if you add a small comment though 😛)

Sources/Bot/API/RepositoryAPI.swift
dmcrodrigues: Good catch 👍

Sources/Bot/Models/CommitState.swift
dmcrodrigues: That's what you get when you copy-paste something from the API without actually processing it 🤦‍♂

Tests/BotTests/GitHub/GitHubAPITests.swift
dmcrodrigues: You're right we don't have them but I'll create an issue to add them but I'd like to not extend this one even more as this can be handled separately.

Tests/BotTests/GitHub/ResponseTests.swift
dmcrodrigues: I think it simplifies a bit 👍

Tests/BotTests/GitHub/GitHubAPITests.swift
dmcrodrigues: 👍

Tests/BotTests/Interceptor/Coordinator.swift
dmcrodrigues: Very good point 👍 I'll include a flag to ensure this gets disabled by default leading to a failure over hitting the real API.

Tests/BotTests/Interceptor/Coordinator.swift
dmcrodrigues: This should be improved in the near future but I didn't want to extend this PR even more in terms of time and complexity, I can add an issue to ensure we track this after being merged.

Tests/BotTests/Mocks/MockGitHubAPI.swift
dmcrodrigues: > Last but not least, not sure if Atomic is enough there. I mean, if 2 requests are run concurrently, you might end up incrementing your iteration and getting index 0 for the first, then for the second, but then the thread/runloop on which first is ran is paused while request 2 is handled, so you end up returning stub 1 before stub 0, right? At least, not totally sure if that would happen that way but my point is that it might be worth not only wrapping the index increment in an atomic operation, but even the whole access to stubs?

I'm missing the problem here because the only thing that we want to protect is the order by which the stubs are requested by, their execution is not controlled in here as we only return `SignalProducer`s which are then started by the consumer so even if we wrap the whole access to `stubs` I don't see how that could prevent or guarantee the correct execution of each producer.

AliSoftware: Yeah I'm not sure there's really a problem/risk with only protecting the counter. I think I was just confused with that part of the code `+1`/`-1` mainly, and the API of `Atomic` (still not sure why they designed it to be both inout and returning)

dmcrodrigues:
@AliSoftware I think I addressed all your feedback, there's just two that I suggest creating an issue over making the changes in this PR once it gets merged.

dmcrodrigues:
We didn't migrate to Swift 5 yet so we can't really use raw strings, I'm reverting that change.

Tests/BotTests/Mocks/MockGitHubAPI.swift
dmcrodrigues: @AliSoftware it provides extra flexibility having that option, as an example you could try to modify something in your value and either succeed or fail, with this option you can represent that by returning a `Bool` to notify that fact.

Another example is the change that we have done, instead of returning just an index we can return the stub directly and do everything behind the scenes.

```swift
private func nextStub() -> Stubs {
         return stubCount.modify { stubCount -> Stubs in
             let nextStub = stubs[stubCount]
             stubCount = stubCount + 1
             return nextStub
         }
}
```

AliSoftware: Thanks makes way more sense with the new code returning a Stub on how it can be useful indeed 👍

dmcrodrigues:
@AliSoftware any chance of reviewing this? Would be great to move it forward.

AliSoftware:
Lool I was already reading it as we speak when you sent your comment 😜

Sources/Bot/API/Decoding.swift
```diff
@@ -0,0 +1,34 @@
...
+    let formatter = DateFormatter()
+    formatter.locale = Locale(identifier:"en_US_POSIX")
+    formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss'Z'"
+    formatter.timeZone = TimeZone(abbreviation:"UTC")
+    return formatter
+}()
+
+func decode<T: Decodable>(_ response: Response) -> Result<T, GitHubClient.Error> {
+    return decode(response) {
+        JSONDecoder.with(dateDecodingStrategy: .formatted(iso8601))
```
AliSoftware: [supernit] newline before `.with`?

Sources/Bot/API/GitHubAPIProtocol.swift
```diff
@@ -1,21 +1,30 @@
...
+    /// - SeeAlso: https://developer.github.com/v3/repos/merging/
+    func merge(head: PullRequest.Branch, into base: PullRequest.Branch) -> SignalProducer<MergeResult, AnyError>
 
-    func mergePullRequest(_ pullRequest: PullRequest) -> SignalProducer<(), AnyError>
+    /// Merges an open pull request.
+    ///
+    /// - Note: This mimics the `Merge Button` from GitHub UI.
+    ///
+    /// - SeeAlso: https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button
+    func mergePullRequest(_ pullRequest: PullRequest) -> SignalProducer<Void, AnyError>
```
AliSoftware, line 23: Great comments 👍

Sources/Bot/API/RepositoryAPI.swift
```diff
@@ -0,0 +1,56 @@
...
+
+    public func postComment(_ comment: String, in pullRequest: PullRequest) -> SignalProducer<Void, AnyError> {
+        return client.request(repository.publish(comment: comment, in: pullRequest))
+            .mapError(AnyError.init)
+    }
+
+    public func removeLabel(_ label: PullRequest.Label, from pullRequest: PullRequest) -> SignalProducer<Void, AnyError> {
+        return client.request(repository.removeLabel(label: label, from: pullRequest))
+            .mapError(AnyError.init)
+    }
```
AliSoftware, line 55: All those methods being all similar scream codegen to me 😜 
(You just woke my codegen instincts, but just joking here, don't worry I'm not saying we should use CodeGen as it's not worth it here, it's just a shame that all those methods are basically the same structure and we have to copy/paste here 😉but not much we can do, we won't bring Sourcery for so little)

Sources/Bot/Extensions/JSONDecoderExtensions.swift
```diff
@@ -0,0 +1,17 @@
...
+    static func with(dateDecodingStrategy strategy: DateDecodingStrategy) -> JSONDecoder {
+        let decoder = JSONDecoder()
+        decoder.dateDecodingStrategy = strategy
+        return decoder
+    }
+
+    func decode<T: Decodable>(_ payload: Data) -> Result<T, DecodingError> {
+        return Result(catching: { () -> T in
+            try decode(T.self, from: payload)
+        })
```
AliSoftware, line 15: [nit] trailing closure syntax would seem fit to me here, but 🤷‍♂ 

```swift
        return Result {
            try decode(T.self, from: payload)
        }
```

Sources/Bot/Services/MergeService.swift
```diff
@@ -111,13 +111,13 @@ extension MergeService {
                         case 0:
                             return github.postComment(
                                 "Your pull request was accepted and is going to be handled right away 🏎",
```
AliSoftware, line 113: Just wording, but since we saw some race conditions in the past in Steve, where 2 PRs were having that message posted in the same time but one of them were in fact waiting in the queue...
```suggestion
                                 "Your pull request was accepted and is going to be handled very soon 🏎",
```

Tests/BotTests/GitHub/GitHubAPITests.swift
AliSoftware: Just asking to be sure you created the ticket (nit: if so it can be nice to add a `// [CNSMR-xxx] Add test for failure cases` comment somewhere here to reference it btw)

Tests/BotTests/Interceptor/Coordinator.swift
```diff
@@ -0,0 +1,99 @@
...
+            return config
+        }())
+    }
+}
+
+final class Coordinator {
+    private var stubs: [Interceptor.Stub] = []
+    private var recordingContext: RecordingContext? = nil
+
+    var isRecordingAvailable: Bool = false
```
AliSoftware, line 23: So can you explain how that works, we're supposed to change that to `true` when we want to switch to record mode, and then back to `false` before pushing to Git (Similar to `self.recordMode = true` on our snapshot tests)?

But that still present the risk of accidentally committing that flag with value `true` if we forget to change it back after recording, right?

Maybe have `Interceptor.stopRecording()` return a boolean, returning `true` if it was actually recording and `false` if it wasn't recording for example, then we could call `XCTestFail("Recording mode on")` in [here](https://github.com/Babylonpartners/Wall-E/pull/14/files#diff-7aab95dc1a106f9b6906895d60608e41R283) when we call `stopRecording` from in `perform(...)` to fail any XCTest when recording is on and ensure the CI fails even if we forgot to turn `isRecordingAvailable` back to `false` after recording.

Tests/BotTests/Interceptor/Coordinator.swift
AliSoftware: Could you add a comment mentioning the ticket you created so we don't forget to implement it?
```suggestion
        // TODO: [CNSMR-???] Check against `URLRequest` to see if we're pop-ing the expected stub
        return stubs.remove(at: 0)
```

Tests/BotTests/Mocks/MockGitHubAPI.swift
```diff
@@ -41,108 +35,74 @@ struct MockGitHubAPI: GitHubAPIProtocol {
...
         }
     }
+
+    private func nextStub() -> Stubs {
+        return stubCount.modify { stubCount -> Stubs in
+            let nextStub = stubs[stubCount]
+            stubCount = stubCount + 1
+            return nextStub
+        }
+    }
```
AliSoftware, line 106: 👌

AliSoftware (approved):
Overall ok 🙂 

You can ignore my small nits if you want, but I'm still not really fan of the protection to avoid recording (and hitting the real API) happening on CI by accident though, so I think we should still fix that. But if you prefer to also make that change in a separate PR, it's fine by me (as long as we track and reference it in comment so we don't forget)

Sources/Bot/API/Decoding.swift
dmcrodrigues: 👍

Sources/Bot/API/RepositoryAPI.swift
dmcrodrigues: Once we remove the `AnyError` I think we can avoid most of this logic.

Sources/Bot/Extensions/JSONDecoderExtensions.swift
dmcrodrigues: That leads to an ambiguous compile error between `init(attempt:)` and `init(catching:)`.

Sources/Bot/Services/MergeService.swift
dmcrodrigues: This is going to be handled separately.

Sources/Bot/Extensions/JSONDecoderExtensions.swift
AliSoftware: Oh, the deprecated one, right? Damn it 😞 Can't wait until we migrate to Swift 5 and use the official one 😅

Tests/BotTests/Interceptor/Coordinator.swift
dmcrodrigues: @AliSoftware good point but what do you think about adding a `fatalError` at the end of `stopRecording` method to abort execution when we are recording instead? Returning a boolean works too but doesn't ensure that the caller will check it/won't ignore it.

```swift
fatalError("Record mode is on. Turn record mode off and re-run your test against the newly-recorded stub.")
```

AliSoftware: I'm not sure but I think `fatalError` might alt all the tests? Feels quite abrupt and would also prevent any cleanup potentially done after the call of `stopRecording` to be executed, be it in the end of the caller's `tearDown` implementation if that's where they call `stopRecording`, or even worse in the middle of the test if one wanted to write a test in such a manner that `stopRecording` would be called in the middle of a test case and not at the very very end...

Wondering though if we could instead do something like:
```swift
#if canImport(XCTest)
XCTestFail(...)
#endif
```
directly inside `stopRecording`'s impl?

dmcrodrigues: Addressed in https://github.com/Babylonpartners/Wall-E/pull/14/commits/d51e27ad563f84410cdbaa597e872b653086c6b1.

Tests/BotTests/GitHub/GitHubAPITests.swift
```diff
@@ -0,0 +1,287 @@
+import XCTest
+import Nimble
+@testable import Bot
+
+class GitHubAPITests: XCTestCase {
+
+    var directory: URL {
+        return URL(string: #file)!
```
mluisbrown, line 8: Curious why you don't use `init(fileURLWithPath: String)` ? With just a string you don't get the `file:///` prefix.

Tests/BotTests/Interceptor/Coordinator.swift
AliSoftware: I trust you tested it in practice and it indeed does the test fail when in record mode?

Tests/BotTests/GitHub/GitHubAPITests.swift
mluisbrown: Ok, I see you are passing this to `FileManager.default.contents(atPath: url.absoluteString)`, which doesn't need the `file:///` prefix 👍

dmcrodrigues: I had that first but was failing while reading, probably I need to use a different API. I can re-test if necessary.

Tests/BotTests/Interceptor/Coordinator.swift
```diff
@@ -0,0 +1,110 @@
...
+    }
+}
+
+#if canImport(XCTest)
+import XCTest
+func fail(_ message: String, file: StaticString = #file, line: UInt = #line) {
+    XCTFail(message)
+}
+#else
+func fail(_ message: String, file: StaticString = #file, line: UInt = #line) {}
```
AliSoftware, line 109: Open discussion, not sure what's best: should we `fatalError` in that `#else` case or not? (probably not now that I think about it, but just wondered what other people think)

Tests/BotTests/Interceptor/Interceptor.swift
```diff
@@ -0,0 +1,55 @@
+import Foundation
+
+private let coordinator = Coordinator()
+
+final class Interceptor: URLProtocol {
```
mluisbrown, line 5: 37wjyg

mluisbrown (approved):
🤖

Tests/BotTests/Interceptor/Coordinator.swift
dmcrodrigues: I did and works as expected, the only problem is that you don't get an indication where it failed because you have a couple of layers between the assertion and the failure but fails and ensures this won't touch the API in CI as we wanted.

Sources/Bot/API/Response.swift
```diff
@@ -0,0 +1,46 @@
+
+// Based on https://github.com/mdiep/Tentacle
+
+import Foundation
+
+let linksRegex = try! NSRegularExpression(pattern: "(?<=\\A|,) *<([^>]+)>( *; *\\w+ *= *\"[^\"]+\")* *(?=\\z|,)")
+let linkParamRegex = try! NSRegularExpression(pattern: "; *(\\w+) *= *\"([^\"]+)\"")
+
+/// Returns any links, keyed by `rel`, from the RFC 5988 link header.
```
ilyapuchka, line 9: reminds me of old days =)
https://github.com/ilyapuchka/Asana/blob/master/Asana/Data%20Access%20Layer/WebLinking.swift#L13

dmcrodrigues: Some things never change 😄

Tests/BotTests/Interceptor/Coordinator.swift
dmcrodrigues: This is only used in the tests so the else clause won't run anyway. I think this can be improved if we decide to use it elsewhere but for now I don't see the point.

Sources/Bot/Services/MergeService.swift
ilyapuchka: managing expectations FTW

ilyapuchka (approved):

Sources/Bot/API/Response.swift
AliSoftware: I think I'll need to borrow that code for Stevenson at some point for multi-page requests to GitHub 😉  (That RFC5988 is a mess to parse, I still don't understand why they haven't made the standard easier to process, but 🤷‍♂)

Sources/Bot/Services/MergeService.swift
dmcrodrigues: @AliSoftware @ilyapuchka fixed in https://github.com/Babylonpartners/Wall-E/pull/21.

PR #18, opened 2019-08-14
[CNSMR-2254] Complete the transition to make Wall-E ready to be used

dmcrodrigues:
### Description

This includes the integration of `GitHubEventsService` in `MergeService` to start consuming all events and handle any pull requests labelled to be merged.

The remaining configuration needed was also included and should be configured as environment variables.

- `GITHUB_TOKEN`: GitHub API Token
- `GITHUB_ORGANIZATION`: the name of the organization where the targeted repository is hosted
- `GITHUB_REPOSITORY`: the name of the targeted repository
- `MERGE_LABEL`: the name of the label used to flag pull requests needed to be merged

### DIFF
--- a/Sources/App/configure.swift
+++ b/Sources/App/configure.swift
@@ -42,34 +42,4 @@ private func makeMergeService(with logger: LoggerProtocol, _ gitHubEventsService
     )
 }
 
-extension Environment {
-
-    static func gitHubWebhookSecret() throws -> String {
-        return try Environment.get("GITHUB_WEBHOOK_SECRET")
-    }
-
-    static func gitHubToken() throws -> String {
-        return try Environment.get("GITHUB_TOKEN")
-    }
-
-    static func gitHubOrganization() throws -> String {
-        return try Environment.get("GITHUB_ORGANIZATION")
-    }
-
-    static func gitHubRepository() throws -> String {
-        return try Environment.get("GITHUB_REPOSITORY")
-    }
-
-    static func mergeLabel() throws -> PullRequest.Label {
-        return PullRequest.Label(name: try Environment.get("MERGE_LABEL"))
-    }
-
-    static func get(_ key: String) throws -> String {
-        guard let value = Environment.get(key) as String?
-            else { throw ConfigurationError.missingConfiguration(message: "💥 key `\(key)` not found in environment") }
-
-        return value
-    }
-}
-
 extension MergeService: Service {}
--- a/Sources/App/configure.swift
+++ b/Sources/App/configure.swift
@@ -9,17 +9,15 @@ enum ConfigurationError: Error {
 public func configure(_ config: inout Config, _ env: inout Environment, _ services: inout Services) throws {
 
     let logger = PrintLogger()
+    let gitHubEventsService = GitHubEventsService(signatureToken: try Environment.gitHubWebhookSecret())
 
     logger.log("👟 Starting up...")
 
-    guard let githubWebhookSecret = Environment.githubWebhookSecret
-        else { throw ConfigurationError.missingConfiguration(message: "💥 GitHub Webhook Secret is missing")}
-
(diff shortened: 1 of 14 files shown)


Sources/App/configure.swift
```diff
@@ -30,3 +28,48 @@ public func configure(_ config: inout Config, _ env: inout Environment, _ servic
...
+        return try Environment.get("GITHUB_TOKEN")
+    }
+
+    static func gitHubOrganization() throws -> String {
+        return try Environment.get("GITHUB_ORGANIZATION")
+    }
+
+    static func gitHubRepository() throws -> String {
+        return try Environment.get("GITHUB_REPOSITORY")
+    }
```
AliSoftware: Nitpicking here but wondering if it wouldn't be more readable to namespace those 4 into a `github` enum to have `Environment.github.webhookSecret`. (Or maybe that's the strange capitalization of `gitHub` that throws me off and just using `githubWebhookSecret()` etc would be enough to make it nicer?)

Sources/App/configure.swift
```diff
@@ -30,3 +28,48 @@ public func configure(_ config: inout Config, _ env: inout Environment, _ servic
...
+
+    return MergeService(
+        integrationLabel: try Environment.mergeLabel(),
+        logger: logger,
+        gitHubAPI: gitHubAPI,
+        gitHubEvents: gitHubEventsService
+    )
+}
+
+extension Environment {
```
AliSoftware: split in separate file?

inamiy (approved):

AliSoftware (approved):
Seems good, mainly assembling existing pieces together 👍 

PS: It's not something we need just right now. but do you have in mind how this would scale in the future if/when we want to support multiple repos (especially maybe Android could be interested for the bot to watch their repo too), should we prefix the env variables with `IOS_` or something – in preparation for the addition of `ANDROID_` repos and labels in the foreseeable future?

dmcrodrigues:
> PS: It's not something we need just right now. but do you have in mind how this would scale in the future if/when we want to support multiple repos (especially maybe Android could be interested for the bot to watch their repo too), should we prefix the env variables with IOS_ or something – in preparation for the addition of ANDROID_ repos and labels in the foreseeable future?

This was briefly discussed and having the same instance managing more than one repository makes everything harder and more complex, I'm not sure if that would be the best option over configuring a separate instance per repository, which was the agreement that we reached.

We can discuss it again but right now we don't have any capacity to accommodate that requirement and its implications.

Sources/App/configure.swift
dmcrodrigues: They are used here so I don't see the advantage in moving them into a separate file, what would be the gain?

Sources/App/configure.swift
dmcrodrigues: They are only 4 and used here once, do we really need to make this more complex/elaborate than it is?

AliSoftware: Probably not. This is just my OCD not liking `gitHub` capitalization in practice. Which is why I suggested to just use `githubRepository()` casing at least... but it's just a matter of taste, so 🤷‍♂

Sources/App/configure.swift
AliSoftware: I'd say convention, organisation and discoverability (if we want to find the env vars we define in the app, and thus the `static func` defined on `Environment`, I'd likely search for `Environment+WallE.swift` or something than having to know it's in `configure.swift` (and not `app.swift` or any other file) 🤷‍♂

dmcrodrigues: @AliSoftware done in https://github.com/Babylonpartners/Wall-E/pull/18/commits/5f51b7aa570f4d18a3ed2354c66684e066e191ae.

AliSoftware: Thanks for appeasing my OCD 😅

PR #19, opened 2019-08-14
Fix wrong path while merging a pull request

dmcrodrigues:
### Description

Our testing revealed this issue which needs to be fixed. We probably need to rethink the strategy to ensure all paths for a certain repository are correctly prefixed as the current system cannot guarantee that.

This also removes a log message which is too verbose at this point.


karabatov (approved):

diegopetrucci (approved):

inamiy (approved):

diegopetrucci:
Also I feel the need to say that the red shade for the `bug` label is very pretty

AliSoftware (approved):

PR #21, opened 2019-08-15
Fix wrong reporting while adding pull requests

dmcrodrigues:
### Description

The reporting that informs the position of the pull request in the queue was not taking in consideration if we have a pull request being integrated already, which means it won't be in the queue anymore, and in the cases where the queue it's empty we were considering the new pull request as the next one to be picked up immediately instead of reporting as the 1st in the queue.

There's a test that checks for this behaviour in specific.

#### Observed behaviour
```
- Status: integrating PR
- Queue: []

Effect ▶️ Adding new PR

*Report*: `Your pull request was accepted and is going to be handled right away 🏎`
```

#### Expected behaviour
```
- Status: integrating PR
- Queue: []

Effect ▶️ Adding new PR

*Report*: `Your pull request was accepted and it's currently `#1` in the queue, hold tight ⏳`
```

### DIFF
--- a/Sources/Bot/Services/MergeService.swift
+++ b/Sources/Bot/Services/MergeService.swift
@@ -120,21 +120,18 @@ extension MergeService {
                         guard previous.pullRequests.firstIndex(of: pullRequest) == nil
                             else { return .empty }
 
-                        switch index {
-                        case 0:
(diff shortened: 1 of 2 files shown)


AliSoftware (approved):
Nice catch!

diegopetrucci (approved):

PR #22, opened 2019-08-15
Attempt of using `QueueScheduler.main` to ensure the correct execution of the bot

dmcrodrigues:
### Description

It seems there might be a problem in terms of QoS causing a delay in the execution/processing of the bot, using the main thread should help us understand if the problem is there or not.

This also cleanups some log messages.

### DIFF
--- a/Sources/App/configure.swift
+++ b/Sources/App/configure.swift
@@ -1,5 +1,6 @@
 import Bot
 import Vapor
+import ReactiveSwift
 
 enum ConfigurationError: Error {
     case missingConfiguration(message: String)
@@ -9,7 +10,10 @@ enum ConfigurationError: Error {
 public func configure(_ config: inout Config, _ env: inout Environment, _ services: inout Services) throws {
 
     let logger = PrintLogger()
-    let gitHubEventsService = GitHubEventsService(signatureToken: try Environment.gitHubWebhookSecret())
+    let gitHubEventsService = GitHubEventsService(
+        signatureToken: try Environment.gitHubWebhookSecret(),
+        scheduler: QueueScheduler.main
+    )
 
     logger.log("👟 Starting up...")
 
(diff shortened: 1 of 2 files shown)


Sources/Bot/Services/MergeService.swift
```diff
@@ -51,7 +51,7 @@ public final class MergeService {
         state.producer
             .combinePrevious()
             .startWithValues { old, new in
-                logger.log("♻️ [Merge Service] Did change state\n - 📜 \(old) \n - 📄 \(new)")
+                logger.log("♻️ Did change state\n - 📜 \(old) \n - 📄 \(new)")
```
AliSoftware, line 54: Why this cleanup? Feels useful in order to do a quick search when using `heroku logs` to find those messages though 😉

Sources/App/configure.swift
```diff
@@ -9,7 +10,10 @@ enum ConfigurationError: Error {
 public func configure(_ config: inout Config, _ env: inout Environment, _ services: inout Services) throws {
 
     let logger = PrintLogger()
-    let gitHubEventsService = GitHubEventsService(signatureToken: try Environment.gitHubWebhookSecret())
+    let gitHubEventsService = GitHubEventsService(
+        signatureToken: try Environment.gitHubWebhookSecret(),
+        scheduler: QueueScheduler.main
```
AliSoftware, line 15: The `QueueScheduler.main` always schedule asynchronously, right? (so no risk of deadlocks from that right?)

Also could a solution to investigate that be to run `Wall-E` locally on your Mac and profile it with Instruments to understand what's happening there?

AliSoftware (approved):

Sources/Bot/Services/MergeService.swift
dmcrodrigues: This bot only manages the merge of pull requests now so this kind of filtering is no longer relevant as we don't have noisy neighbours anymore 😃

Sources/App/configure.swift
dmcrodrigues: Running it locally is the next step if this doesn't work but that would require some configuration and point the webhook to my machine, there's no easy way to debug this end-to-end.

Sources/Bot/Services/MergeService.swift
AliSoftware: Ah, good point!

RuiAAPeres (approved):

PR #23, opened 2019-10-03
[IOSP-169] Upgrade to Swift 5

### LINKED ISSUE #17: Migrate to Swift 5
We are still based on Swift 4.2.1 and we should make the migration to Swift 5.

dmcrodrigues:
Ticket: https://babylonpartners.atlassian.net/browse/IOSP-169

Closes #17.


Tests/BotTests/GitHub/GitHubAPITests.swift
```diff
@@ -2,6 +2,18 @@ import XCTest
 import Nimble
 @testable import Bot
 
+extension Result {
+
+    var value: Success? {
```
AliSoftware, line 10: [nit] This can also be achieved directly with the built-in `try? result.get()` instead of `result.value`, so just wondering if that extension is really necessary? 🤔 (but hey if it's more readable at call site and only extended for tests anyway, it's ok to keep it)

AliSoftware (approved):

.circleci/config.yml
```diff
@@ -19,16 +19,28 @@ references:
...
   test:
     docker:
-      - image: swift:4.2
+       - image: vapor/swift:5.1
     steps:
       - checkout
       - run:
           name: Compile code
-          command: swift build
+          command: swift build --jobs 1
```
AliSoftware, line 27: Note to self: we might want the same for Stevenson, to avoid random failures we sometimes have during build of Vapor (which a re-run usually fixes)

Sources/Bot/API/Response.swift
```diff
@@ -30,7 +30,7 @@ private func linksInLinkHeader(_ header: String) -> [String: URL] {
 }
 
 /// A response from the GitHub API.
-internal struct Response: Hashable {
+public struct Response: Hashable {
```
AliSoftware, line 33: [nit] should we rename that to `GitHubResponse` or something to avoid ambiguity with `Vapor.Response`? (could be done in separate PR, just a nit)

AliSoftware (approved):
Looking good, let's finally merge this 👍

Sources/Bot/API/Response.swift
dmcrodrigues: Nevertheless the comment this type is generic enough to be used in other environments but it's something we can do yes 👍

ilyapuchka (approved):

Sources/Bot/API/Response.swift
dmcrodrigues: Let's do it in a separate PR Olivier.

PR #24, opened 2019-10-10
[IOSP-177] Define a healthcheck route to access the health of the bot

dmcrodrigues:
To deploy Wall-E we need a healthcheck route to assess the health of the service. However, this can be particularly tricky to do because all network activity is signed so we cannot use a `ping` event for example.

Based on that, we are monitoring the internal state of `MergeService` to monitor if the state keeps changing, meaning there's activity and PRs are being picked up and merged until it finally changes back to `idle`. This can be done with the concept of timeouts, because we have a specific time window to handle a PR, based on `statusChecksTimeout`, we can estimate that by then the state should have changed already if it haven't means there's something wrong and unexpected. This system is not super strong but it's a bit hard to monitor this concept.

### DIFF
--- a/Sources/Bot/Services/MergeService.swift
+++ b/Sources/Bot/Services/MergeService.swift
@@ -16,6 +16,8 @@ public final class MergeService {
     private let statusChecksCompletion: Signal<StatusEvent, NoError>
     private let statusChecksCompletionObserver: Signal<StatusEvent, NoError>.Observer
 
+    public let healthcheck: Healthcheck
+
     public init(
         integrationLabel: PullRequest.Label,
         statusChecksTimeout: TimeInterval = 60.minutes,
@@ -47,6 +49,8 @@ public final class MergeService {
                 Feedbacks.whenAddingPullRequests(github: self.gitHubAPI, scheduler: scheduler)
             ]
         )
+
+        healthcheck = Healthcheck(state: state.signal, statusChecksTimeout: statusChecksTimeout, scheduler: scheduler)
 
         state.producer
             .combinePrevious()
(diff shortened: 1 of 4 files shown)


Sources/Bot/Services/MergeService.swift
```diff
@@ -401,6 +405,45 @@ extension MergeService {
...
+            case unhealthy(Reason)
+        }
+
+        public let status: Property<Status>
+
+        internal init(
+            state: Signal<State, NoError>,
+            statusChecksTimeout: TimeInterval,
+            scheduler: DateScheduler
+            ) {
```
AliSoftware:
```suggestion
        ) {
```

Sources/Bot/Services/MergeService.swift
```diff
@@ -401,6 +405,45 @@ extension MergeService {
...
+            status = Property(
+                initial: .ok,
+                then: state.combinePrevious()
+                    .skipRepeats { lhs, rhs in
+                        return lhs == rhs
+                    }
```
AliSoftware, line 432: I'm not a RAS expert, but wouldn't `state.skipRepeats()` be enough there instead of `state.combinePrevious().skipRepeats { … == … }`? I mean I thought `skipRepeats` would do that for you already

AliSoftware (approved):

Sources/Bot/Services/MergeService.swift
```diff
@@ -401,6 +405,45 @@ extension MergeService {
...
+                    .flatMap(.latest) { _, current -> SignalProducer<Status, NoError> in
+                        switch current.status {
+                        case .starting, .idle:
+                            return SignalProducer(value: .ok)
+                        default:
+                            return SignalProducer(value: .unhealthy(.potentialDeadlock))
+                                .delay(1.5 * statusChecksTimeout, on: scheduler)
```
ilyapuchka, line 446: can we add comment here explaining this delay?

ilyapuchka (approved):
Please add comment in code about delay and maybe what this "health check" means 🙏

viorel15:
[NIT]initially I thought this is related to our Healthcheck services form the app. 
IMO SanityCheck sounds better.

viorel15 (approved):

Sources/Bot/Services/MergeService.swift
dmcrodrigues: We are working with tuples here and in Swift 4.2 they not `Equatable` out of the box.

AliSoftware: Oh 😢

dmcrodrigues:
_health check_ is a common term in the world of (micro-)services and we will be deployed using Kubernetes with active monitoring to assess the health of the service.

Sources/Bot/Services/MergeService.swift
dmcrodrigues: Included @ilyapuchka, let me know if it's clear enough.

ilyapuchka: a lil bit complex but ok =)

dmcrodrigues: Yeah it's kind of hard to set-up a logic for this 😕

PR #25, opened 2019-10-10
Add a docker image to run Wall-E as a container

dmcrodrigues:
This is a base image to run Wall-E as a container. The image itself still needs to be deployed and hosted in either Docker Hub or [quay.io](https://quay.io).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM swift:4.2
+
+WORKDIR /Wall-E
+COPY . /Wall-E
+
+RUN swift build -c release
+
+EXPOSE 2008
+
+CMD ./.build/release/Run --env production --hostname 0.0.0.0 --port 2008


Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM swift:4.2
+
+WORKDIR /Wall-E
+COPY . /Wall-E
+
+RUN swift build -c release
+
+EXPOSE 2008
```
AliSoftware, line 8: Curious as to why this specific port, is there any reason? (so that I know if I have to do a similar thing in the future)

Did you had to ask DevOps to open a port for the future instance and Wall-E and they just opened this one and told you the port number they'd be opening for Wall-E specifically?

AliSoftware (approved):
I don't have enough experience in Docker to validate if there might be something missing in this `Dockerfile`, but from what I just read from the Docker documentation, it seems to make sense and use the right commands, so LGTM

ilyapuchka (approved):

Dockerfile
dmcrodrigues: @AliSoftware 😛

image

But this is not signed-off yet.

AliSoftware: Ohhh nice easter-egg 😄

PR #26, opened 2019-10-16
[IOSP-207] Support Top Priority PRs

AliSoftware:
https://babylonpartners.atlassian.net/browse/IOSP-207

### Why?

As a simple way to allow us making some PRs jump at top of the queue – until we implement a more complex logic in [IOSP-164](https://babylonpartners.atlassian.net/browse/IOSP-164) later

### How?

* `MergeService` must now be initialized with a `topPriorityLabels` list in addition to the `integrationLabel` string.
* When adding or removing a Label on a PR, the logic is still the same (addition/removal of a label leads to `Outcome.include` or `Outcome.exclude` depending on presence of the `integrationLabel`) but if integrationLabel is still there, the `Outcome.include` will lead to a queue update instead of filtered addition

### TODO

We need to add the `TOP_PRIORITY_LABELS` environment variable to Heroku after this

### Notes

* To reorder the queue, we couldn't use `sort` nor `partition(by:)` as originally thought, because they are not guaranteed to be stable sort / stable partition, so we'd risk not maintaining the original relative order
* the `XCTestManifests.swift` file didn't seem to have been updated since a while. We might want to create a ticket to add a job on a Mac container in the future that would run `swift test --generate-manifest` and assert there's no git changes afterwards


Sources/App/Extensions/EnvironmentProperties.swift
```diff
@@ -23,6 +23,14 @@ extension Environment {
         return PullRequest.Label(name: try Environment.get("MERGE_LABEL"))
     }
 
+    // TODO: OHA: Add this value to env vars in host
```
AliSoftware, line 26: To remove once heroku/AWS config has been updated to include
`TOP_PRIORITY_LABELS=Feature at risk 🚑,Hotfix 🆘`

```suggestion
```

AliSoftware:
Got a PullAssigner error. Will assign manually for now but we should look into it.

image

[EDIT] Added the other GH team to the repo to solve it. At some point we need to consolidate those two teams into one (or make it explicit what the diff is if there's a need to keep the two separate)

Tests/BotTests/MergeServiceHealthcheckTests.swift
```diff
@@ -8,33 +8,43 @@ class MergeServiceHealthcheckTests: XCTestCase {
 
     private let statusChecksTimeout = 30.minutes
 
+    private func makeState(status: MergeService.State.Status) -> MergeService.State {
+        return .init(
+            integrationLabel: .init(name: "Merge"),
```
zzcgumn: I would prefer that we move these definitions to a fixture, perhaps `LabelFixture`, and pass values defined in that fixture as default arguments to the `makeState` function. I think it is very likely that we will want customise our test cases with different values.

(The other changes in this file makes it very clear why introducing this factory method is a good idea :) )

Sources/Bot/Services/MergeService.swift
```diff
@@ -677,7 +709,8 @@ extension MergeService.State: CustomDebugStringConvertible {
         let pullRequestsSeparator = "\n\t\t"
 
         let pullRequestsRepresentation = pullRequests.enumerated().map { index, pullRequest in
-            return "#\(index + 1): \(pullRequest)"
+            let isTP = pullRequest.isLabelled(withOneOf: self.topPriorityLabels)
+            return "#\(index + 1): \(pullRequest) \(isTP ? "[TP]" : "")"
```
zzcgumn, line 713: This is very debatable in my view. Do we really want to rely on hardcoded pseudo code strings?

Tests/BotTests/MergeServiceTests.swift
```diff
@@ -673,6 +674,111 @@ class MergeServiceTests: XCTestCase {
         )
     }
 
+    func test_changing_pull_request_priorities() {
```
zzcgumn, line 663: Thanks for adding these tests. Gives much better confidence.

Sources/Bot/Services/MergeService.swift
AliSoftware: Not sure I'm following. This is only used to get a `debugDescription` of the queue of `MergeService` for the logs, what do you mean by "hardcoded _pseudo-code_ strings" exactly 🤔?

zzcgumn: My mistake. I misread the code and thought they were part of the business logic. All good then.

Tests/BotTests/MergeServiceTests.swift
AliSoftware: I actually found a small bug in my algorithm while I was adding them, so it reassured me too 😉

Sources/Bot/Services/MergeService.swift
```diff
@@ -506,37 +508,64 @@ extension MergeService {
...
+                topPriorityLabels: topPriorityLabels,
+                statusChecksTimeout: statusChecksTimeout,
+                pullRequests: pullRequests,
+                status: status
+            )
         }
 
         func include(pullRequests pullRequestsToInclude: [PullRequest]) -> State {
+            let onlyNewPRs = pullRequestsToInclude.filter {
+                [enqueued = self.pullRequests.map { $0.number }] pullRequest in
```
ilyapuchka: imo having such expression in a capture list is kind of confusing. maybe just capture `self.pullRequests` and do `map().contains()`

ilyapuchka (approved):

Tests/BotTests/MergeServiceHealthcheckTests.swift
AliSoftware: @zzcgumn Suggestion applied in latest commits

Sources/Bot/Services/MergeService.swift
AliSoftware: Done in 17b23cc

zzcgumn (approved):
LGTM

PR #28, opened 2019-10-28
[IOSP-178] Update CI to build and deploy the docker image

dmcrodrigues:
We want to build and deploy the docker image in our CI so this includes both processes. Deploying only happens after tagging a new release.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,16 +1,20 @@
 version: 2.1
 
-- &quay_login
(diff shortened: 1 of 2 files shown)


AliSoftware (approved):
I don't know much about `Docker` so since I don't see anything flagrant out of place and commands seems to match what we previously had in heroku, I trust you checked with DevOps and will approve 😉

ilyapuchka (approved):
👍

zzcgumn (approved):

PR #30, opened 2019-10-29
Simplify deployments

dmcrodrigues:
Using multiple jobs for build and deploy will require the usage of workspaces to share data across them, at this point we don't need such an elaborate workflow so we can have a single job responsible for deployment. Building the docker image as part of each PR is also not something relevant at this point.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,11 +15,6 @@ references:
         only:
         - /.*/
 
-commands:
-  quay_login:
-    steps:
-      - run: docker login -u "${DOCKER_USER}" -p "${DOCKER_PASS}" quay.io
-
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -33,38 +28,31 @@ jobs:
...
+          command: docker login -u "${DOCKER_USER}" -p "${DOCKER_PASS}" quay.io
       - run:
           name: Deploy image
           command: make install
 
 workflows:
   version: 2
 
-  test-build-deploy:
+  test-n-deploy:
```
AliSoftware, line 50: rock-n-roll! 🎸

AliSoftware (approved):

j531 (approved):

ilyapuchka (approved):

PR #31, opened 2019-10-29
[IOSP-196] Include config to only consider *required* checks to merge

AliSoftware:
Ticket: https://babylonpartners.atlassian.net/browse/IOSP-196

## Why?

At the moment Wall-E checks for **all** the checks part of the PR, and refuses to merge it if any of those has failed.

This means there's the assumption that alls checks must pass, but that is not always the case, as some teams have some informative checks included that not always pass, and being informative checks they are not marked as required.

## How?

Wall-E now accepts an environment variable to configure if all checks need to pass or only required checks are required to pass to merge the pull request.

(This is specially needed to enable Wall-E to be used by the Android team)

### DIFF
--- a/Sources/App/Extensions/EnvironmentProperties.swift
+++ b/Sources/App/Extensions/EnvironmentProperties.swift
@@ -19,6 +19,13 @@ extension Environment {
         return try Environment.get("GITHUB_REPOSITORY")
     }
 
+    static func requiresAllGitHubStatusChecks() throws -> Bool {
+        guard let stringValue: String = try? Environment.get("REQUIRES_ALL_STATUS_CHECKS") else {
+            return false // defaults to only consider required checks
+        }
+        return ["yes", "1", "true"].contains(stringValue.lowercased())
+    }
+
     static func mergeLabel() throws -> PullRequest.Label {
         return PullRequest.Label(name: try Environment.get("MERGE_LABEL"))
     }
--- a/Sources/Bot/Models/RequiredStatusCheck.swift
+++ b/Sources/Bot/Models/RequiredStatusCheck.swift
@@ -1,0 +1,6 @@
+import Foundation
+
+public struct RequiredStatusChecks: Decodable, Equatable {
+    public let strict: Bool // Require branches to be up to date before merging?
+    public let contexts: [String] // Names of status checks marked as Required
+}
--- a/Sources/Bot/Models/RequiredStatusCheck.swift
+++ b/Sources/Bot/Models/RequiredStatusCheck.swift
@@ -2,7 +2,12 @@ import Foundation
 
 public struct RequiredStatusChecks: Decodable, Equatable {
     /// Require branches to be up to date before merging?
-    public let strict: Bool
+    public let isStrict: Bool
     /// Names of status checks marked as Required
     public let contexts: [String]
+
+    enum CodingKeys: String, CodingKey {
+        case isStrict = "strict"
+        case contexts
+    }
 }
--- a/Sources/Bot/Models/RequiredStatusCheck.swift
+++ b/Sources/Bot/Models/RequiredStatusCheck.swift
@@ -1,6 +1,8 @@
 import Foundation
 
 public struct RequiredStatusChecks: Decodable, Equatable {
-    public let strict: Bool // Require branches to be up to date before merging?
-    public let contexts: [String] // Names of status checks marked as Required
+    /// Require branches to be up to date before merging?
+    public let strict: Bool
+    /// Names of status checks marked as Required
+    public let contexts: [String]
 }
--- a/Sources/Bot/Services/MergeService.swift
+++ b/Sources/Bot/Services/MergeService.swift
@@ -417,8 +417,15 @@ extension MergeService {
         }
     }
 
-    fileprivate static func getRequiredChecksState(github: GitHubAPIProtocol, targetBranch: PullRequest.Branch, commitState: CommitState) -> SignalProducer<CommitState.State, AnyError> {
-        github.fetchRequiredStatusChecks(for: targetBranch).map { (requiredStatusChecks) -> CommitState.State in
+    fileprivate static func getRequiredChecksState(
+        github: GitHubAPIProtocol,
+        targetBranch: PullRequest.Branch,
+        commitState: CommitState
+    ) -> SignalProducer<CommitState.State, AnyError> {
+        guard commitState.state != .success else {
+            return .value(.success)
+        }
+        return github.fetchRequiredStatusChecks(for: targetBranch).map { (requiredStatusChecks) -> CommitState.State in
             let requiredStates = commitState.statuses.filter { status in
                 requiredStatusChecks.contexts.contains(status.context)
             }
--- a/Sources/Bot/Services/MergeService.swift
+++ b/Sources/Bot/Services/MergeService.swift
@@ -21,6 +21,7 @@ public final class MergeService {
     public init(
         integrationLabel: PullRequest.Label,
         topPriorityLabels: [PullRequest.Label],
+        requiresAllStatusChecks: Bool,
         statusChecksTimeout: TimeInterval = 60.minutes,
         logger: LoggerProtocol,
         gitHubAPI: GitHubAPIProtocol,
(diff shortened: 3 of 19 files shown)


Sources/Bot/Services/MergeService.swift
```diff
@@ -234,7 +237,7 @@ extension MergeService {
                         case .pending:
                             return .value(.integrationDidChangeStatus(.updating, metadata))
                         case .failure:
-                            return .value(.integrationDidChangeStatus(.failed(.checkingCommitChecksFailed), metadata))
+                            return .value(.integrationDidChangeStatus(.failed(.checksFailing), metadata))
```
AliSoftware, line 240: This was a typo I discovered while working on the codebase, figured I'd fix it while I was at it

Sources/Bot/Services/MergeService.swift
```diff
@@ -402,6 +412,31 @@ extension MergeService {
...
+    /// i.e. returns .failure or .pending if one of the required check is in .failure or .pending respectively
+    /// and return .success only if all required states are .success
+    fileprivate static func getRequiredChecksState(
+        github: GitHubAPIProtocol,
+        targetBranch: PullRequest.Branch,
+        commitState: CommitState
+    ) -> SignalProducer<CommitState.State, AnyError> {
+        if commitState.state == .success {
+            return .value(.success)
+        }
```
AliSoftware, line 426: this is because the `CommitState.state` property is already `.success` only if all checks (required or otherwise) are `.success`. So if all checks (required or otherwise) are successful, no need to make the API call to fetch the list of required checks on the target branch

Tests/BotTests/MergeService/MergeServiceTests.swift
```diff
@@ -427,45 +404,41 @@ class MergeServiceTests: XCTestCase {
...
                 .postComment { _, _ in },
                 .mergeIntoBranch { _, _ in .success },
                 .getPullRequest { _ in MergeServiceFixture.defaultTarget.with(mergeState: .blocked) },
-                .getCommitStatus { _ in CommitState(state: .pending, statuses: []) },
+                .getCommitStatus { _ in CommitState.stub(states: [.pending]) },
+                .getRequiredStatusChecks{ _ in requiredStatusChecks },
                 .getPullRequest { _ in MergeServiceFixture.defaultTarget.with(mergeState: .blocked) },
-                .getCommitStatus { _ in CommitState(state: .pending, statuses: []) },
+                .getCommitStatus { _ in CommitState.stub(states: [.pending]) },
+                .getRequiredStatusChecks{ _ in requiredStatusChecks },
```
AliSoftware, line 439: ideally we might want to cache the result of `getRequiredStatusChecks` to avoid API calls being repeated in the future. otoh if we happen to have changed the Status Checks (e.g. we updated the CircleCI config and the job names) between the 2 calls, if we add cache we could have some strange caching issues.

I've decided to keep the decision on implementing caching for this API for a later PR (if ever, since we might change the strategy of relying on undocumented `MergeState` in the future anyway, for which we'll then need to make more API calls and may want to refactor that part even more in those cases, which will lead to a broader ticket anyway)

Sources/App/Extensions/EnvironmentProperties.swift
```diff
@@ -19,6 +19,13 @@ extension Environment {
         return try Environment.get("GITHUB_REPOSITORY")
     }
 
+    static func requiresAllGitHubStatusChecks() throws -> Bool {
+        guard let stringValue: String = try? Environment.get("REQUIRES_ALL_STATUS_CHECKS") else {
+            return false // defaults to only consider required checks
+        }
+        return ["yes", "1", "true"].contains(stringValue.lowercased())
```
ilyapuchka, line 26: doesn't `Bool(init: StringLiteral)` do it already?

Sources/Bot/Models/RequiredStatusCheck.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+
+public struct RequiredStatusChecks: Decodable, Equatable {
+    public let strict: Bool // Require branches to be up to date before merging?
```
ilyapuchka: these comments would be better as proper doc comments as these properties are public.

Sources/Bot/Models/RequiredStatusCheck.swift
```diff
@@ -0,0 +1,6 @@
...
+    public let contexts: [String] // Names of status checks marked as Required
```
ilyapuchka: why name it `contexts`? If this is based on Github API names maybe better to hide these details into CodingKeys?

Sources/Bot/Services/MergeService.swift
```diff
@@ -190,7 +192,8 @@ extension MergeService {
                 else { return .value(.integrationDidChangeStatus(.done, metadata)) }
 
             switch metadata.mergeState {
-            case .clean:
+            case .clean,
+                 .unstable where !requiresAllStatusChecks:
```
ilyapuchka, line 196: what does `unstable` mean?

Tests/BotTests/MergeService/Stubs/CommitState+Stub.swift
```diff
@@ -0,0 +1,35 @@
+@testable import Bot
+
+extension CommitState {
+    static func stub(states: [CommitState.State]) -> CommitState {
+        guard !states.isEmpty else {
+            return CommitState(state: .pending, statuses: [])
+        }
+
+        let combinedState: CommitState.State
```
ilyapuchka: if we are stubbing commit state to behave like this, why do we need the same logic in the implementation at https://github.com/babylonhealth/Wall-E/pull/31/files#diff-39d0b4d1a105d5e23d5b7b2556fb51fdR431? Won't github commit status property follow the same rules? If not then why are we stubbing it like it does?

ilyapuchka (changes requested):
Overall looks good but some questions first

Sources/App/Extensions/EnvironmentProperties.swift
AliSoftware: Does that exist? TIL
Though:
1. We're not supposed to use `init(stringLiteral:)` initialisers ourselves nor with non-literal values, those are supposed to be used by compiler only officially
2. I didn't find any `Bool.init(stringLiteral: )`  in the documentation?
3. There's `Bool.init?(_: String)` but the doc for this one says:

> If the description value is any string other than "true" or "false", the result is nil. This initializer is case sensitive.

Which seems a bit too strict to me, especially since by convention most ENV vars that expect a bool value in most contexts expect either `YES` or `1` most of the time, not lowercase `true`

Sources/Bot/Models/RequiredStatusCheck.swift
AliSoftware: I thought about this, but the term `context` is already used in the existing codebase for those terms in multiple places. I could indeed rename them here, but in that case better rename those everywhere, including in the `CommitStatus` and `StatusEvent` model etc too (which both existed prior to this PR).

I think it's better to keep the GitHub API naming to keep consistency.

Sources/Bot/Services/MergeService.swift
AliSoftware: This is documented as a proper doc-comment on the `enum case` itself (i.e. you can alt-click on it to see quick help).

```swift
        /// Failing/pending commit status that is not part of the required status checks. Merging is still allowed.
        case unstable
```

Tests/BotTests/MergeService/Stubs/CommitState+Stub.swift
AliSoftware: github commit status property only follow that rule but for _all_ the checks. While the https://github.com/babylonhealth/Wall-E/pull/31/files#diff-39d0b4d1a105d5e23d5b7b2556fb51fdR431 implementation uses a similar rule… but ONLY for _required_ checks.

So those implementations are similar but different in practice, as this stub will stub the behavior of the GitHub API but the one in `fetchRequiredStatusChecks` will only consider the required checks (which is exactly why we need `getRequiredChecksState` in the first place)

ilyapuchka: can you add these details about `commit.status` to the doc comment of a `getRequiredChecksState` method?

AliSoftware: Just extracted it in a dedicated method with proper doc comment in a69dfbb03f5e169cc12dfea8f5eb989ed502f6b0 .

image

Sources/Bot/Models/RequiredStatusCheck.swift
AliSoftware: Fixed in a69dfbb03f5e169cc12dfea8f5eb989ed502f6b0

ilyapuchka (approved):

Sources/Bot/Models/RequiredStatusCheck.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+public struct RequiredStatusChecks: Decodable, Equatable {
+    /// Require branches to be up to date before merging?
+    public let strict: Bool
```
inamiy:
```suggestion
    public let isStrict: Bool
```

[Style-Guide](https://github.com/babylonhealth/ios-playbook/blob/master/Cookbook/Style-guide/README.md)

> boolean types should read like assertions (isLoading, isHidden, isFalse, etc)

Sources/Bot/Models/RequiredStatusCheck.swift
```diff
@@ -0,0 +1,8 @@
...
+    /// Names of status checks marked as Required
+    public let contexts: [String]
```
inamiy, line 7: Definition (Doc-comment) of `strict` and `contexts` look quite overlapping to me,
so we might want to simplify `struct RequiredStatusChecks` , possibly by using enum instead.

dmcrodrigues (approved):

Sources/Bot/Models/RequiredStatusCheck.swift
dmcrodrigues: They are two independent concepts, branches can have a list of checks marked as required that must pass to enable merging but they can also have an option to force PRs to be up to date with the base branch and they are not mutually exclusive, you can have just one of them enabled, both or even none.

inamiy: OK I was a bit confused due to these params order.

`contexts` should probably come first as a required param, and `strict` comes secondary as a possibly optional param (which can add default value e.g. `true`), so changing order will improve readability in this case.

If we read `strict` first, `contexts` also seem strict required params that overlaps previous keyword.

Sources/Bot/Models/RequiredStatusCheck.swift
inamiy: Done in 25a6004

PR #35, opened 2019-11-04
Using the main scheduler leads to a deadlock

dmcrodrigues:
Ticket: https://babylonpartners.atlassian.net/browse/IOSP-293

Unfortunately #22 caused more harm than good and introduced a deadlock. This can be verified in `MergeService` which doesn't log any change of state while we were expecting at least going from `idle` to `ready`.

### DIFF
--- a/Sources/App/configure.swift
+++ b/Sources/App/configure.swift
@@ -1,6 +1,5 @@
 import Bot
 import Vapor
-import ReactiveSwift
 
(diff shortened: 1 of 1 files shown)


ilyapuchka (approved):

mluisbrown (approved):

NSMyself (approved):
🚢

zzcgumn (approved):

PR #36, opened 2019-11-05
[IOSP-299] Include a route to inform about the current state of the bot

dmcrodrigues:
Ticket: https://babylonpartners.atlassian.net/browse/IOSP-299

This includes a new route to return the state of the queue as plain text. It will return something along this lines.

```
State(
 - status: idle,
 - queue: []
)
```

The PR in the process of being integrated will be part of `status` and `queue` will contain all PRs waiting to be merged.

### DIFF
--- a/Sources/App/routes.swift
+++ b/Sources/App/routes.swift
@@ -8,6 +8,10 @@ public func routes(
     gitHubEventsService: GitHubEventsService
 ) throws {
 
+    router.get("/") { request -> String in
+        return String(describing: mergeService.state.value)
+    }
+
     router.get("health") { request -> HTTPResponse in
         switch mergeService.healthcheck.status.value {
         case .ok: return HTTPResponse(status: .ok)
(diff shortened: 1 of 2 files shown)


Sources/App/routes.swift
```diff
@@ -8,6 +8,10 @@ public func routes(
     gitHubEventsService: GitHubEventsService
 ) throws {
 
+    router.get("/") { request -> String in
```
ilyapuchka, line 11: maybe better to do it `/status`?

ilyapuchka (approved):

Sources/App/routes.swift
dmcrodrigues: Initially I was using that route but maybe using the direct URL for the bot is easier to remember? 🤷‍♂️

ilyapuchka: true

NSMyself (approved):
🚢

danspndl (approved):
🚀

mluisbrown (approved):

PR #39, opened 2019-11-22
[IOSP-164] Multiple Merge Queues (one per target branch)

AliSoftware:
Ticket: https://babylonpartners.atlassian.net/browse/IOSP-164

This PR allows the bot to have multiple Merge queues in parallel, one per target branch, so that we can in particular have PRs targeting `release/*` being merged in parallel to PRs targeting `develop`

Note: This PR also updates WallEView, which is the Client App introduced very recently in #41 which relies on the queue description output (`/` route) and since now this endpoint returns the list of queues for all MergeServices instead of a single queue, changing the API of that endpoint, we use that PR to update the parsing of that endpoint in the ClientApp too


AliSoftware:
@dmcrodrigues I still need to add some tests on `DispatchService` but the preliminary implementation should be ready to have an overall review.

(PS: CI fails because of Swift versions mismatch, I'm on it)

Tests/BotTests/MergeService/MergeServiceTests.swift
```diff
@@ -970,35 +1003,69 @@ class MergeServiceTests: XCTestCase {
...
+                case .destroyed(let service):
+                    statesObserver.send(value: .mergeServiceDestroyed(forBranch: service.targetBranch))
+                }
+        }
 
         when(gitHubEvents, scheduler)
-        assert(states)
+
+        statesObserver.sendCompleted()
+        states.collect().observeValues(assert)
```
AliSoftware: Note: I'm not a fan of this (async) approach that uses `collect()` and an `@escaping` closure, but it seams that other approaches I've tried led to race conditions and thus an inconsistent order in the states that got collected.

For example I tried using an approach similar to `TestObserver` by having `var states: [DispatchServiceState] = []` then using `statesObserver.observeValues { states.append($0) }` and finally assert on `states` values – like [in this commit here](b8d5d847bf5ca3e8e91db52b69959bef8dd84438), but it led to the test failing, claiming that it got an order ending with `[…, runningStatusChecks, .idle, .ready]` – which is not possible because `reduceRunningStatusChecks` can't go to `.idle`, only to `.ready` (and only transition possible to `.idle` is from `.starting` or `.ready`), so clearly the issue was with the way of collecting state values…

Not sure how it can be that `.ready` and `.idle` could be swapped like that: it would probably mean that they're emitted one right after the other, during the same "heartbeat" of the Scheduler, but given that `.idle` state is supposed to only transition from `ready` using RAF after a `.noMorePullRequests` event, they should be emitted one after the other anyway, right?

Any suggestion to improve that part is welcome.

Sources/App/routes.swift
```diff
@@ -4,22 +4,22 @@ import Vapor
...
-    mergeService: MergeService,
+    dispatchService: DispatchService,
     gitHubEventsService: GitHubEventsService
 ) throws {
 
     router.get("/") { request -> Response in
         let response = Response(using: request)
         if request.header(named: HTTPHeaderName.accept.description) == "application/json" {
-            try response.content.encode(mergeService.state.value, as: .json)
+            try response.content.encode(dispatchService.queues, as: .json)
```
AliSoftware: @ilyapuchka Please note that this will change the API used by the ClientApp you recently introduced, now returning [a dictionary of `[String: MergeState.State]` indexed by the branch names](https://github.com/babylonhealth/Wall-E/pull/39/files#diff-eaa970022ee17cbc1398a964b5d4a44eR147-R149) – since we'll now have multiple Merge Queues in parallel – instead of what was previously just a single `MergeService.State`.

ilyapuchka: Is `queues` property used for anything else? If not then it would be easier for paring to make `branch` a property of `MergeService.state` and encode in into json

AliSoftware: Nope `queues` is only used there – in fact I only introduced it in the merge commit when I tried to merge `master` into this branch and had to fix conflicts with your change about JSON queue, so it has introduced just for that.

AliSoftware: Fixed in 820623f

Sources/Bot/Services/DispatchService.swift
```diff
@@ -0,0 +1,198 @@
+import Foundation
+import Result
+import ReactiveSwift
+import ReactiveFeedback
+
+public final class DispatchService {
```
ilyapuchka, line 7:
```suggestion
/// Orchestrates multiple merge services, one per each target branch of PRs enqueued for integration
public final class DispatchService {
```

Sources/Bot/Services/DispatchService.swift
```diff
@@ -0,0 +1,198 @@
...
+    private let integrationLabel: PullRequest.Label
+    private let topPriorityLabels: [PullRequest.Label]
+    private let requiresAllStatusChecks: Bool
+    private let statusChecksTimeout: TimeInterval
+
+    private let logger: LoggerProtocol
+    private let gitHubAPI: GitHubAPIProtocol
+    private let scheduler: DateScheduler
+
+    private var mergeServices: [String: MergeService]
```
ilyapuchka:
```suggestion
    /// Merge services per target branch
    private var mergeServices: [String: MergeService]
```

Sources/Bot/Models/PullRequest.swift
```diff
@@ -70,3 +70,12 @@ extension PullRequest.Branch: CustomDebugStringConvertible {
         return "Branch(\(ref), \(sha))"
     }
 }
+
+extension PullRequest {
+    func isLabelled(with label: PullRequest.Label) -> Bool {
+        return labels.contains(label)
+    }
+    func isLabelled(withOneOf possibleLabels: [PullRequest.Label]) -> Bool {
```
ilyapuchka, line 78: is this being used? couldn't find it

Sources/Bot/Services/DispatchService.swift
```diff
@@ -0,0 +1,198 @@
...
+                case let .status(event):
+                    self?.statusChecksDidChange(event: event)
+                case .ping:
+                    break
+                }
+            }
+    }
+
+    private func dispatchInitial(pullRequests: [PullRequest]) {
+        let dispatchTable = Dictionary(grouping: pullRequests) { pullRequest in pullRequest.target.ref }
```
ilyapuchka: [nit] even though we prefer not to use unnamed parameters, in this case it's quite obvious that each item of `pullRequests` is `pullRequest`, so I'd shorten it

```suggestion
        let dispatchTable = Dictionary(grouping: pullRequests) { $0.target.ref }
```

Sources/Bot/Services/DispatchService.swift
```diff
@@ -0,0 +1,198 @@
...
+        for (branch, pullRequestsForBranch) in dispatchTable {
+            makeMergeService(targetBranch: branch, scheduler: self.scheduler, initialPullRequests: pullRequestsForBranch)
```
ilyapuchka:
```suggestion
            makeMergeService(
                targetBranch: branch,
                scheduler: self.scheduler,
                initialPullRequests: pullRequestsForBranch
            )
```

Sources/Bot/Services/DispatchService.swift
```diff
@@ -0,0 +1,198 @@
...
+        let targetBranch = event.pullRequestMetadata.reference.target.ref
+        if let service = mergeServices[targetBranch] {
+            service.pullRequestChangesObserver.send(value: (event.pullRequestMetadata, event.action))
+        } else {
+            makeMergeService(targetBranch: targetBranch, scheduler: self.scheduler, initialPullRequests: [event.pullRequestMetadata.reference])
+        }
+    }
+
+    private func statusChecksDidChange(event: StatusEvent) {
+        // No way to know which MergeService this event is supposed to be for – isRelative(toBranch:) only checks for head branch not target so not useful here
```
ilyapuchka, line 104: so there is no target branch in the PR metadata with the event? 🤔

Sources/Bot/Services/DispatchService.swift
```diff
@@ -0,0 +1,198 @@
...
+            requiresAllStatusChecks: requiresAllStatusChecks,
+            statusChecksTimeout: statusChecksTimeout,
+            initialPullRequests: initialPullRequests,
+            logger: logger,
+            gitHubAPI: gitHubAPI,
+            scheduler: scheduler
+        )
+        self.mergeServices[targetBranch] = mergeService
+        self.healthcheck.startMonitoring(mergeServiceHealthcheck: mergeService.healthcheck)
+        mergeServiceLifecycleObserver?.send(value: .created(mergeService))
```
ilyapuchka: maybe we should remove this out of this function so that this side effect is outside of the function?

Sources/Bot/Services/DispatchService.swift
```diff
@@ -0,0 +1,198 @@
...
+        mergeService.state.producer
```
ilyapuchka, line 126: this observation as well maybe? it's just a bit weird that we do everything here, return the value, but then never use it (as far as I can tell)

we can then extract this observations into method, like `observerMergeServiceLifetime` or something like that

Sources/Bot/Services/MergeService.swift
```diff
@@ -147,24 +321,22 @@ extension MergeService {
...
     fileprivate static func whenReady(github: GitHubAPIProtocol, scheduler: Scheduler) -> Feedback<State, Event> {
         return Feedback(predicate: { $0.status == .ready }) { state -> SignalProducer<Event, NoError> in
 
-            guard let next = state.pullRequests.first
-                else { return .value(.noMorePullRequests) }
+            guard let next = state.pullRequests.first else {
+                return SignalProducer
+                    .value(.noMorePullRequests)
+                    .observe(on: scheduler)
+            }
```
AliSoftware, line 338: This missing `.observe(on: scheduler)` is not related to my changes directly but was the one causing the race condition on my tests (but only on Linux, so that was not easy to debug). Took a while to find it 😅

Sources/Bot/Services/DispatchService.swift
```diff
@@ -0,0 +1,198 @@
...
+            .observe(on: scheduler)
+            .startWithValues { [weak self, service = mergeService] state in
+                self?.mergeServiceLifecycleObserver?.send(value: .stateChanged(service))
+                if state.status == .idle {
+                    self?.mergeServices[targetBranch] = nil
+                    self?.mergeServiceLifecycleObserver?.send(value: .destroyed(service))
```
ilyapuchka: another idea is to call these `.send` methods in the setter of `mergeServices`? it might be a bit tricky though to find what key actually changed...

Sources/Bot/Services/DispatchService.swift
```diff
@@ -0,0 +1,198 @@
...
+            producers[uuid] = mergeServiceHealthcheck.status.producer
+                .on(disposed: { [weak self] in
+                    self?.producers[uuid] = nil
+                })
+
+            statusProducerDisposable?.dispose()
+            statusProducerDisposable = SignalProducer
+                .combineLatest(producers.values)
+                .map({ (mergeServiceStatuses: [MergeService.Healthcheck.Status]) -> Healthcheck.Status in
+                    return mergeServiceStatuses.contains(where: { $0 != .ok }) ? .unhealthy(.potentialDeadlock) : .ok
```
ilyapuchka:
```suggestion
                   mergeServiceStatuses.contains(where: { $0 != .ok }) ? .unhealthy(.potentialDeadlock) : .ok
```

Sources/Bot/Services/MergeService.swift
```diff
@@ -751,18 +695,56 @@ extension MergeService.State {
...
+
+        internal init(
+            state: Signal<State, NoError>,
+            statusChecksTimeout: TimeInterval,
+            scheduler: DateScheduler
+        ) {
+            status = Property(
+                initial: .ok,
+                then: state.combinePrevious()
+                    .skipRepeats { lhs, rhs in
```
ilyapuchka, line 721: `skipRepeats()` should be enough no?

Sources/Bot/Services/DispatchService.swift
```diff
@@ -0,0 +1,198 @@
...
+            """
+            ## Merge Queue for target branch: \(entry.key) ##
+
+            \(entry.value.state.value)
+            """
+        }.joined(separator: "\n\n")
+    }
+
+    public var queueStates: [MergeService.State] {
+        return self.mergeServices.values.map { $0.state.value }
```
ilyapuchka: should we maybe define coding keys for state manually so that we don't encode things from the state that we don't need, like labels?

Sources/Bot/Models/PullRequest.swift
AliSoftware: It's used in line 230 of `MergeService.swift`, in order to reorder the queue for PRs being labelled with one of the topPriority labels.

(It's not really new code from this PR, it just got moved around to reorganize code)

Sources/Bot/Services/DispatchService.swift
AliSoftware: If I understand [the doc](https://developer.github.com/v3/activity/events/types/#statusevent) correctly, only the list of branches this check _affects_, so the _source_ branches, are part of the payload, but not target branches?

Sources/Bot/Services/DispatchService.swift
AliSoftware: Not sure where would you move it to though? At what point in the code would you trigger this instead?

AliSoftware: Or were you just suggesting to just extract that in a new private function and call that private function from here?

ilyapuchka: not from here but from the loop that calls this function:

```
for ... {
  let mergeService = makeMergeService(...)
  observeMergeServiceLifetime(mergeService)
}
```

not critical though, just surprised by the discardable result

Sources/Bot/Services/DispatchService.swift
AliSoftware: Yeah I thought about that but it felt more complicated that it was worth to have to do the diff manually and detect new keys instead of directly trigger them at the point we delete them ourselves – `mergeServices` is private after all so nothing other that this code should alter it.

I can explore the solution about observing `mergeServices` changes but that would require us to transform it to a `Property` instead of a single dictionary, just to be able to observe it and `combinePrevious` and do some math, and I'm seeing more risks of doing that wrong than benefits tbh, no?

Sources/Bot/Services/MergeService.swift
AliSoftware: I remember asking that to @dmcrodrigues  too back in the day when I saw this code (which is not new, was just moved in this PR from one place to another), and he gave me a good justification for it at that point, but I admit I can't remember the rationale 😅 (which means it would be worth a comment once we remember the reason again)

Sources/Bot/Services/DispatchService.swift
AliSoftware: Well I though you defined those manually in your PR about JSON output already? 🤔 

But you're right, I forgot to add the `targetBranch` as part of the keys since you customized the CodingKeys in your PR (see lines 768-779 of `MergeService.swift`), I should add this new `targetBranch` key indeed.

ilyapuchka: I guess we are both right)

Sources/Bot/Services/DispatchService.swift
AliSoftware: Ah right, I see what you mean.
But since we want to observe lifecycle _every time_ we create a new MergeService via `makeMergeService` it feels safer to add it inside the `makeMergeService` rather than making sure to always call `observeMergeServiceLifecycle` every time we call `makeMergeService`?

Sources/Bot/Services/MergeService.swift
dmcrodrigues: Still using Swift 4.2 and because we applied `combinePrevious` we are dealing with tuples that do not conform to `Equatable`.

Sources/Bot/Services/DispatchService.swift
ilyapuchka: yes, it seems more complicated this way

ilyapuchka:
@AliSoftware since client app PR was merged, do you mind if we add a small fix to decoding on client app side in this PR so that we don't break it?

Sources/Bot/Services/MergeService.swift
AliSoftware: Comment added in 128eb64cceb642a8d9492fa5a5b082aa6d23c897 about this.

Sources/Bot/Services/DispatchService.swift
ilyapuchka: not sure, as in example there are multiple branches. Can we validate that? Maybe using our fork of the repo?

AliSoftware: Well on line 486 of `MergeService.swift` we do:

```swift
            return statusChecksCompletion
                .observe(on: scheduler)
                .filter { change in change.state != .pending && change.isRelative(toBranch: pullRequest.source.ref) }
                .on { change in
                    logger.log("📣 Status check `\(change.context)` finished with result: `\(change.state)` (SHA: `\(change.sha)`)")
                }
```
Which means we use that `isRelative(to:)` method to check against… the `source` branch (not the target branch), which should confirm my understanding at least…

And since the `MergeService` do filter those status checks (which it can do because it has the reference to all its PRs which reference said source branch so can use that to filter, as opposed to `DispatchService` which only know the `targetBranch` of each `MergeService`), not sure that really matters anyway 😉

Sources/Bot/Services/DispatchService.swift
AliSoftware: Fixed + added relevant unit tests in 309bdfecf3d3245fb15abc60b96540d9f0561baf

AliSoftware:
> since client app PR was merged, do you mind if we add a small fix to decoding on client app side in this PR so that we don't break it?

@ilyapuchka Sure; I'll let you push the required changes then though 😉

Sources/Bot/Services/DispatchService.swift
```diff
@@ -0,0 +1,208 @@
...
+    public init(
+        integrationLabel: PullRequest.Label,
+        topPriorityLabels: [PullRequest.Label],
+        requiresAllStatusChecks: Bool,
+        statusChecksTimeout: TimeInterval,
+        logger: LoggerProtocol,
+        gitHubAPI: GitHubAPIProtocol,
+        gitHubEvents: GitHubEventsServiceProtocol,
+        scheduler: DateScheduler = QueueScheduler(),
+        mergeServiceLifecycleObserver: Signal<MergeServiceLifecycleEvent, NoError>.Observer? = nil
```
dmcrodrigues: This doesn't need to be injected and defined by the caller, it's a bit strange to have it that way to be honest, `DispatchService` can create a signal and expose to be consumed by any interested parties which in this case will be `DispatchServiceContext`.

So this moves from `DispatchServiceContext` to `DispatchService` which keeps the observer private, as we don't want any other parties sending those events other than `DispatchService` itself and exposes the signal to keep the observation in `DispatchServiceContext`.

`let (lifecycleSignal, lifecycleObserver) = Signal<DispatchService.MergeServiceLifecycleEvent, NoError>.pipe()`

AliSoftware: Good point. That implementation comes from multiple iterations when I worked on the testing and I didn't properly pause and look at the big picture once it ended up working like I needed. Will refactor to implement as you suggested, makes more sense to me indeed.

Sources/Bot/Services/DispatchService.swift
```diff
@@ -0,0 +1,208 @@
...
+    private let topPriorityLabels: [PullRequest.Label]
+    private let requiresAllStatusChecks: Bool
+    private let statusChecksTimeout: TimeInterval
+
+    private let logger: LoggerProtocol
+    private let gitHubAPI: GitHubAPIProtocol
+    private let scheduler: DateScheduler
+
+    /// Merge services per target branch
+    private var mergeServices: [String: MergeService]
```
dmcrodrigues: This needs to be thread-safe so we need to wrap it using `Atomic` and using the proper methods to access and mutate it, we may have PRs coming while one of the merge services changes to a `idle` state.

Sources/Bot/Services/DispatchService.swift
```diff
@@ -0,0 +1,208 @@
...
+
+        public enum Status: Equatable {
+            case ok
+            case unhealthy(Reason)
+        }
+
+        public var status: Property<Status> { return Property(_status) }
+
+        private let scheduler: DateScheduler
+        private var producers: [String: SignalProducer<MergeService.Healthcheck.Status, NoError>] = [:]
```
dmcrodrigues: This also needs to be thread-safe. However, considering the current usage of this root health check, and the underlying implementation which is a bit complex, I think we can live with something more simple like having a function `func currentStatus() -> Status` that internally uses this dictionary and queries each merge service's healthcheck and `reduce`s everything into one consolidated status, what do you think @AliSoftware ?

AliSoftware: Agreed, I was actually not very happy with this implementation and this kinda dirty trick of disposing and recreating the producer, so I'm happy with an on-demand impl instead 👍

Sources/Bot/Services/DispatchService.swift
AliSoftware: Fixed in a34f44573fa7b17d433abc08a60a738e8ad86306

.gitignore
```diff
@@ -3,5 +3,6 @@ Packages
 .swiftpm
 xcuserdata
 *.xcodeproj
+!WallEView.xcodeproj
```
ilyapuchka, line 6: Had to add it so that we don't have to recreate project every time. Not sure if we already can use SPM for application targets?

WallEView/ViewController.swift
```diff
@@ -103,13 +113,14 @@ class ViewController: NSViewController, NSTableViewDataSource, NSTableViewDelega
             DispatchQueue.main.async {
                 if let data = data {
                     do {
-                        let state = try JSONDecoder().decode(State.self, from: data)
-                        self?.state = state
+                        let queues = try JSONDecoder().decode([Queue].self, from: data)
```
ilyapuchka, line 116: @AliSoftware adjusted parsing here, for now just filtering everything that is no targeted to develop branch. In future I think we can add a drop down selected for user to select target branch and to switch between different queues. what do you think?

Sources/Bot/Services/DispatchService.swift
AliSoftware: @dmcrodrigues Should be fixed in 700d62d8e9d2d67c9606d76ae52c218a23e59275 but another pair of eyes to double-check is welcome

.gitignore
AliSoftware: I'm a bit lost with the state of SPM those days, but indeed I'm not sure it's yet possible, given that `.app` bundle contain resources (like at least the plist and the app icon) and those are still not yet supported by SPM (though that'll likely change soon)

WallEView/ViewController.swift
```diff
@@ -39,24 +39,34 @@ class ViewController: NSViewController, NSTableViewDataSource, NSTableViewDelega
...
+        static let empty = State(queues: [])
+    }
+
+    struct Queue {
         struct Current: Decodable {
             let reference: PullRequest
         }
+        let targetBranch: String
         let current: Current?
         let queue: [PullRequest]
```
AliSoftware, line 71: [nit] rename `queue` to `pending` maybe (by opposition to `current`)? (not that important though)

It feels a bit strange and not easy to understand to have a `Queue` type that _contains_… a `queue` property inside: is the type the queue, or is it _containing_ a queue?  🤔feels confusing at first read

WallEView/ViewController.swift
AliSoftware: @ilyapuchka Reading your code it's more about "filtering everything that is not the _first_ queue. Agreed that since most of our target branches are either `develop` or `release/*`, then `develop` is likely to be the first queue (they are ordered alphabetically in the JSON API response), but what if Ben creates a branch `ben/xxx-123` then want to split his work and have a second PR that builds on top of the first one and thus makes a PR from `ben/xxx-123-bis` branch to `ben/xxx-123` branch… then this new target branch will be the `.first` though 😉 

Maybe it's worth adding a convenience method `var developQueue: Queue? { return queues.filter { $0.targetBranch == "develop" }` on `State` and use that instead of `queues.first` throughout your changes here?

.gitignore
AliSoftware: Ok just got confirmation from another Slack, SPM still can't generate macOS bundles so we still need the `xcodeproj` indeed.

WallEView/ViewController.swift
ilyapuchka: I think it's ok to filter just develop for now, as soon as I add branch selector we won't need this code, and we won't need `developQueue` property as well

Sources/Bot/Services/DispatchService.swift
AliSoftware: Fixed in 80e3d6300f654dc5e2881d013c48191cd2542bde
This indeed simplifies the code _a lot_ 👍

Sources/Bot/Services/DispatchService.swift
```diff
@@ -0,0 +1,175 @@
...
+            }
+        }
+    }
+
+    private func pullRequestDidChange(event: PullRequestEvent) {
+        logger.log("📣 Pull Request did change \(event.pullRequestMetadata) with action `\(event.action)`")
+        let targetBranch = event.pullRequestMetadata.reference.target.ref
+        mergeServices.modify { dict in
+            if let service = dict[targetBranch] {
+                service.pullRequestChangesObserver.send(value: (event.pullRequestMetadata, event.action))
```
inamiy: Side-effects inside `Atomic.modify` is not encouraged and may cause deadlock, so it's better just returning `service` here, and then call 

```swift
if let service = mergeServices.modify({ ... }) { 
    /* after unlock, send to observer */ 
    service.pullRequestChangesObserver.send(value: (event.pullRequestMetadata, event.action))
}
```

Sources/Bot/Services/MergeService.swift
```diff
@@ -106,6 +91,195 @@ public final class MergeService {
...
+
+            }
+        }
+
+        internal let targetBranch: String
+        internal let integrationLabel: PullRequest.Label
+        internal let topPriorityLabels: [PullRequest.Label]
+        internal let statusChecksTimeout: TimeInterval
+        public let pullRequests: [PullRequest]
+        public let status: Status
```
inamiy: [nit] Can we bring these `State` members to the top of the scope (as in style-guide) ?

Sources/Bot/Services/MergeService.swift
```diff
@@ -751,18 +695,57 @@ extension MergeService.State {
...
+        public enum Reason: Error, Equatable {
+            case potentialDeadlock
+        }
+
+        public enum Status: Equatable {
+            case ok
+            case unhealthy(Reason)
+        }
+
+        public let status: Property<Status>
```
inamiy: [nit] This is the dominant member variable for `class Healthcheck` so pls move to top of scope.

Tests/BotTests/MergeService/DispatchServiceContext.swift
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    var branch: String {
+        switch self {
+        case .created(let branch), .destroyed(let branch):
+            return branch
+        case .state(let state):
+            return state.targetBranch
+        }
+
```
inamiy:
```suggestion
```

inamiy (commented):
Can you breifly explain about what `WallEView` is for? (in PR description, and ideally, documentation)

AliSoftware:
> Can you breifly explain about what WallEView is for? (in PR description, and ideally, documentation)

PR Description updated.

@inamiy WallEView is not directly related to this PR but is the ClientApp that @ilyapuchka developed in #41. That Client App makes use of the JSON output of the merge queue (#40) and since with this current PR introducing multiple merge queue the JSON got modified to return all the current merge queues instead of a single one, we took the occasion of this PR to update the client app as well. See [this comment above](https://github.com/babylonhealth/Wall-E/pull/39#issuecomment-558620455) on which we agreed that it could be done in that PR to avoid breaking the client app once it was merged.

Sources/Bot/Services/DispatchService.swift
AliSoftware: Fixed in 2f89f2ba609bfa934ae35e70f5ebd27a851be226

Sources/Bot/Services/MergeService.swift
AliSoftware: Fixed in 6345029c29e9bd24a03d4a99c4e9eba56931e4b5

Sources/Bot/Services/MergeService.swift
AliSoftware: Fixed in 6345029c29e9bd24a03d4a99c4e9eba56931e4b5

inamiy (approved):

dmcrodrigues (approved):
Good stuff Olivier 👍 @ilyapuchka can you review this one too before merging?

ilyapuchka:
@dmcrodrigues you know this better, so if you approve I approve. 😉

ilyapuchka (approved):

PR #42, opened 2019-12-02
Implement a bit extended logging

ilyapuchka:
Added function name and line number to error states so that it's a bit simpler to debug

### DIFF
--- a/Sources/App/Extensions/EnvironmentProperties.swift
+++ b/Sources/App/Extensions/EnvironmentProperties.swift
@@ -43,7 +43,7 @@ extension Environment {
         }
     }
 
-    static func extendedLogging() -> Bool? {
+    static func extendedLogging() -> Bool {
         return (try? Environment.get("EXTENDED_LOGGING").bool) ?? false
     }
 
--- a/Sources/App/Extensions/EnvironmentProperties.swift
+++ b/Sources/App/Extensions/EnvironmentProperties.swift
@@ -43,6 +43,10 @@ extension Environment {
         }
     }
 
+    static func extendedLogging() -> Bool? {
+        return (try? Environment.get("EXTENDED_LOGGING").bool) ?? false
+    }
+
     static func get(_ key: String) throws -> String {
         guard let value = Environment.get(key) as String?
             else { throw ConfigurationError.missingConfiguration(message: "💥 key `\(key)` not found in environment") }
--- a/Sources/App/configure.swift
+++ b/Sources/App/configure.swift
@@ -47,7 +47,7 @@ private func makeMergeService(with logger: LoggerProtocol, _ gitHubEventsService
         logger: logger,
         gitHubAPI: gitHubAPI,
         gitHubEvents: gitHubEventsService,
-        extendedLogging: Environment.extendedLogging() ?? false
+        extendedLogging: Environment.extendedLogging()
     )
 }
 
--- a/Sources/App/configure.swift
+++ b/Sources/App/configure.swift
@@ -46,7 +46,8 @@ private func makeMergeService(with logger: LoggerProtocol, _ gitHubEventsService
         statusChecksTimeout: try Environment.statusChecksTimeout() ?? 90.minutes,
         logger: logger,
         gitHubAPI: gitHubAPI,
-        gitHubEvents: gitHubEventsService
+        gitHubEvents: gitHubEventsService,
+        extendedLogging: Environment.extendedLogging() ?? false
     )
 }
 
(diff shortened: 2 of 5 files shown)


Sources/Bot/Services/MergeService.swift
```diff
@@ -353,28 +353,43 @@ extension MergeService {
...
             let integrationLabel: PullRequest.Label
             let failureReason: FailureReason
+            let trace: Trace
+            let extendedLogging: Bool
 
             var failureMessage: String {
-                return "@\(pullRequest.author.login) unfortunately the integration failed with code: `\(failureReason)`."
+                let message = "@\(pullRequest.author.login) unfortunately the integration failed with code: `\(failureReason)`."
+                if extendedLogging {
+                    return message + "\nfunction: \(trace.function) line: \(trace.line)"
```
ilyapuchka:
```suggestion
                    return message + "\n<details>function: \(trace.function) line: \(trace.line)</details>"
```

Sources/Bot/Services/MergeService.swift
```diff
@@ -353,28 +353,43 @@ extension MergeService {
...
             let integrationLabel: PullRequest.Label
             let failureReason: FailureReason
+            let trace: Trace
+            let extendedLogging: Bool
 
             var failureMessage: String {
-                return "@\(pullRequest.author.login) unfortunately the integration failed with code: `\(failureReason)`."
+                let message = "@\(pullRequest.author.login) unfortunately the integration failed with code: `\(failureReason)`."
+                if extendedLogging {
+                    return message + "\n<details>function: \(trace.function) line: \(trace.line)</details>"
```
AliSoftware, line 370: Feels it would be more natural to make `trace` be `CustomStringConvertible` and just use `<details>\(trace)</details>` here

Sources/Bot/Services/MergeService.swift
```diff
@@ -483,6 +498,20 @@ extension MergeService {
...
+        let line: UInt
+
+        static func capture(function: String = #function, line: UInt = #line) -> Trace {
+            return self.init(function: function, line: line)
+        }
+
+        public static func ==(lhs: Trace, rhs: Trace) -> Bool {
+            // we don't really want to compare this in tests
+            return true
+        }
```
AliSoftware, line 512: Then why does it need to be equatable in the first place?

ilyapuchka: for automatic code generation for conformance of enum cases?

AliSoftware: mmmh. Not sure it's worth it, since it could then lead to hard to understand and debug situations in the future in case we end up actually adding tests that do an `==` on two `Trace` objects for some reason…

ilyapuchka: We shouldn't write such tests because they will be breaking on each refactoring with no value in that, and this is purely info for debugging.

AliSoftware: Well sure but with your shortcut, they won't break which they should 😜 Just feels quite weird to have an `Equatable` implementation that doesn't tell the truth is all 😉 🤷‍♂

ilyapuchka: It feels weird to write such tests and to implement equality on whole State just because of this type

Sources/App/Extensions/EnvironmentProperties.swift
```diff
@@ -43,6 +43,10 @@ extension Environment {
         }
     }
 
+    static func extendedLogging() -> Bool? {
```
karabatov:
```suggestion
    static func extendedLogging() -> Bool {
```

Sources/App/configure.swift
```diff
@@ -46,7 +46,8 @@ private func makeMergeService(with logger: LoggerProtocol, _ gitHubEventsService
         statusChecksTimeout: try Environment.statusChecksTimeout() ?? 90.minutes,
         logger: logger,
         gitHubAPI: gitHubAPI,
-        gitHubEvents: gitHubEventsService
+        gitHubEvents: gitHubEventsService,
+        extendedLogging: Environment.extendedLogging() ?? false
```
karabatov: If you declare it as `Bool` then you don't need this…
```suggestion
        extendedLogging: Environment.extendedLogging()
```

Sources/Bot/Services/MergeService.swift
karabatov: Maybe.

karabatov (approved):

PR #43, opened 2019-12-03
[IOSP-436] Check for any pending checks before integrating clean PR

ilyapuchka:
Tested this with https://github.com/babylonhealth/walle-debug/pull/3

One check is configured to run on push and one on API calls (like we have in our repo).
Now when there are pending checks bot waits for them to succeed. We can do that by manually triggering check via API. When it's done bot is merging PR as expected.


Sources/Bot/Services/MergeService.swift
```diff
@@ -402,27 +402,36 @@ extension MergeService {
...
-                            return github.fetchPullRequest(number: metadata.reference.number)
-                                .map { metadata in
-                                    switch metadata.mergeState {
-                                    case .clean:
-                                        return .retryIntegration(metadata)
-                                    default:
-                                        return .integrationDidChangeStatus(.failed(.blocked), metadata)
+                let pullRequest = metadata.reference
+                return github.fetchAllStatusChecks(for: pullRequest).map { statusChecks -> Bool in
+                    return statusChecks.map({ $0.state }).contains(.pending)
```
ilyapuchka, line 409: for some reason github reports checks not as I expect, on ci they are called "test" and "build" but when I tested it this request returned one completed check (one that was run) and two pending checks, all with name "build" 🤷‍♂ 
I mean it works for what we need to do here, but is unexpected

danspndl: I think every CI run is a `build`, there are 2 types of status checks on GH: checks and statuses. Status is the "old" checks, which may support less types. Checks is the tab we tried in the temp repo, it may have more rich features (and probably more CI job types as well)

Sources/Bot/Services/MergeService.swift
```diff
@@ -402,27 +402,36 @@ extension MergeService {
...
+                                        .map { metadata in
+                                            switch metadata.mergeState {
+                                            case .clean:
+                                                return .retryIntegration(metadata)
+                                            default:
+                                                return .integrationDidChangeStatus(.failed(.blocked), metadata)
+                                            }
                                     }
-                            }
+                                }
```
danspndl: [nit]
```suggestion
                            }
```

danspndl (approved):
LGTM 🚢

MichalTKwiecien (approved):

Sources/Bot/Services/MergeService.swift
```diff
@@ -402,27 +402,36 @@ extension MergeService {
...
-                                    default:
-                                        return .integrationDidChangeStatus(.failed(.blocked), metadata)
+                let pullRequest = metadata.reference
+                return github.fetchAllStatusChecks(for: pullRequest).map { statusChecks -> Bool in
+                    return statusChecks.map({ $0.state }).contains(.pending)
+                }.flatMap(.latest) { pendingStatusChecks -> SignalProducer<Event, AnyError> in
+                    if pendingStatusChecks {
+                        return .value(.integrationDidChangeStatus(.updating, metadata))
+                    } else {
+                        return github.fetchCommitStatus(for: metadata.reference)
```
AliSoftware, line 414: Not completely sure here (still need my morning coffee 😄) but couldn't we derive the `switch` below from the `statusChecks` that we fetched in `fetchAllStatusChecks`?
I mean, instead of doing 2 API calls (fetching all status checks, then if none pending, fetch commit status and switch on that), maybe there's a way to only do the `fetchAllStatusChecks` API call and determine everything from there?

(The `static func combinedState(for states: [CommitState.State]) -> CommitState.State` method that we already have in `CommitState.swift` can even help in that)

ilyapuchka: I wouldn't go there considering that what we are using is undocumented APIs and inferring will meant that we will be trying to replicate undocumented logic. I also wanted to make as little changes as possible, so kept the rest of logic the same.

dmcrodrigues: I don't think we can avoid two API calls in this case but we need to re-work this logic a bit because for the failure we need to check if we want all checks to pass or just the required checks and then check if the failure is not in one of them like we do in the running status checks feedback and for that we need the API call to get all required checks. 

I think we missed this part when we introduced the concept of requiring only the required checks to pass.

AliSoftware: You mean `fetchAllStatusChecks` is undocumented API? (if so, worth a comment in code)

I thought the only undocumented API we were using was for the `commitStatus.mergeState: MergeState` that we use after the call to `fetchCommitStatus`+`fetchPullRequest` – not the `CommitState.State` we use in `fetchAllStatusChecks`? 🤔 
So in that sense wouldn't getting rid of the undocumented API for `MergeState` and only relying on `CommitState.State` returned by `fetchAllStatusChecks` be better?

dmcrodrigues: > I wouldn't go there considering that what we are using is undocumented APIs

I don't think that's the case here @ilyapuchka, the only undocumented feature that we are using is the merge state part of the pull request, the API calls and combining the state of each status into a single one is something that we can do.

ilyapuchka: I can look into that but personally I'd prefer to ship this fix faster as we have to keep answering people about this bug fixes that this change and they are affected by it quite ofter, and do optimisations later. I don't see how getting rid of one API call will significantly improve anything here.

dmcrodrigues: We are not talking about getting rid of one API call Ilya but something that's currently lacking to honour the configuration that we have to only check the required checks instead of all of them but we can address that in a separate pull request.

dmcrodrigues: In any case, tests are failing so we cannot ship the fix yet.

ilyapuchka:
I suppose we have some flaky tests then because these tests are not related to the fix and they are passing locally.

dmcrodrigues:
@AliSoftware can you have a look please?

AliSoftware:
> @AliSoftware can you have a look please?

@dmcrodrigues test flakiness should be fixed by latest commit
@ilyapuchka remaining test failures are related to your changes I believe (maybe just missing `scheduler.advance()`?)

ilyapuchka:
@AliSoftware failing tests are all in DispatchServiceTests, I don't think my changes could affect them because otherwise they will crash on missing network stub. They also pass locally for me. Merge with master does not solve it.

ilyapuchka:
I see you added a commit now

ilyapuchka:
The same tests still fail though

AliSoftware:
I added a commit to fix the flakiness of what failed in your original PR, which was the JSON queue comparison (because serialisation on macOS on Linux are different and lead to different key orders)

Once I added that commit, new tests started to fail on CI (but not locally) about the `MergeService`'s list of states. I wouldn't be so quick as to say that because they are in `DispatchServiceTests` your changes are not related to those, because `DispatchServiceTests` test the list of states a `MergeService` can have and that can be because there's one more step in your `MergeService` now (e.g. a new API call requiring an additional `scheduler.advance()` in the tests for example – thinking about this one possibility especially because the diff is just a couple of missing states at the end of the list of expected states).

I'll try to investigate further.

Sources/Bot/Services/MergeService.swift
AliSoftware: > We are not talking about getting rid of one API call Ilya but something that's currently lacking to honour the configuration that we have to only check the required checks instead of all of them but we can address that in a separate pull request.

FWIW [commit 1b486b0](https://github.com/babylonhealth/Wall-E/commit/1b486b0295c9825a14feef2c7388a418d1d49ee4#diff-39d0b4d1a105d5e23d5b7b2556fb51fdR409) in this PR did introduce the fetching of required status checks, not sure that logic was later removed in subsequent commit?

AliSoftware:
@dmcrodrigues @ilyapuchka I've tried to fix the test failure but I'm a bit blocked, couldn't understand what's different on Linux vs macOS and why the `reduceReady` is not called after the `Event.noMorePullRequests` is sent by `Feedback.whenReady` (confirmed by print-debugging on CI like old times…)…

At first I thought this could be because we forgot to `observe(on: scheduler)` on some code and accidentally used the main scheduler instead of the `TestScheduler` during tests… but the more I investigate the more it seems it might actually be an issue inside RAF itself (at least @konrad-em seem to [have had a similar issue in the past too](https://babylonhealth.slack.com/archives/GD6J0QWQ5/p1575556259211500?thread_ts=1575554936.208900&cid=GD6J0QWQ5), and I think he ended up rewriting his code to avoid RAF…)

Any help understanding what could be wrong in RAF Linux or in our logic, and to fix the tests, would be appreciated. I've opened a draft PR here: https://github.com/babylonhealth/Wall-E/pull/47

Sources/Bot/Services/MergeService.swift
ilyapuchka: yes, it was removed as it's not exactly what we want to fetch here.

dmcrodrigues: > it was removed as it's not exactly what we want to fetch here

How do handle the scenario where the configuration only requires required checks to pass and you have one non-required failing? `fetchCommitStatus` will return `failure` and consequently mark the integration has failed even tho users are ok in having optional checks failing.

Tests/BotTests/MergeService/DispatchServiceTests.swift
```diff
@@ -33,6 +33,7 @@ class DispatchServiceTests: XCTestCase {
             ],
             when: { service, scheduler in
                 scheduler.advance()
+                scheduler.advance()
```
dmcrodrigues: What's happening here exactly? This is very strange. cc @AliSoftware

Sources/Bot/Services/MergeService.swift
ilyapuchka: We now only check if any statuses are still pending before checking for commit status as we used to. If that logic is not sufficient then I think it falls into what you mentioned before that we oversaw with requires status checks configuration. My understanding was that this can be tackled separately as it's more than just what this fix addresses (a race condition between triggering checks and checking them), or did I misunderstood?

Tests/BotTests/MergeService/DispatchServiceTests.swift
ilyapuchka: I suppose this is somehow related to the test fixes on Linux, but as they didn't really solved the issue maybe we should remove these changes from this PR?

AliSoftware: Yeah we should remove that. Was a (desperate) attempt to see if that was the reason for the test failure on Linux. But we'll try to handle the debugging of that in #47 so we should indeed remove that from this PR.
```suggestion
```

Sources/Bot/Services/MergeService.swift
dmcrodrigues: But if there's a commit with that logic already in place why did we remove it? 😕 We will need to bring those changes anyway either in this PR or in a following one.

ilyapuchka: As per our discussion in Slack I realised it was not exactly what we needed here, so I changed from getting required checks to getting all checks and only checking if anything is still pending
https://babylonhealth.slack.com/archives/GQBDJ9422/p1575374833013900

dmcrodrigues: Ok let's tackle this in a separate ticket.

Tests/BotTests/MergeService/DispatchServiceTests.swift
ilyapuchka: Ok, I'll revert your commit then @AliSoftware I suppose it's already in another branch dedicated to the tests fix

Sources/Bot/Services/MergeService.swift
AliSoftware: Note that if the bot is configured to ignore failures from non-required GitHub checks, we might consider that we don't care if those non-required GitHub checks are still pending and could consider merging the PR as soon as all required checks are green even if non-required are still orange.

That can be for a separate ticket and PR though, but I think it's worth adding a `// TODO: [IOSP-xxx]` comment in there to not forget.

```suggestion
                    // TODO: [IOSP-???] Only wait for pending checks that are _required_ if bot is configured to ignore non-required checks
                    return statusChecks.map({ $0.state }).contains(.pending)
```

AliSoftware (approved):

PR #44, opened 2019-12-03
View multiple queues

ilyapuchka:
This PR adds a branch selector to be able to view multiple queues

### DIFF
--- a/WallEView/ViewController.swift
+++ b/WallEView/ViewController.swift
@@ -15,6 +15,12 @@ extension NSTableView {
     }
 }
 
+class GridClipTableView: NSTableView {
+    // workaround to hide separators for non populated rows
+    // https://stackoverflow.com/questions/5606796/draw-grid-lines-in-nstableview-only-for-populated-rows
+    override func drawGrid(inClipRect clipRect: NSRect) { }
+}
+
 class PullRequestCell: NSTableCellView {
     @IBOutlet weak var title: NSTextField!
     @IBOutlet weak var subtitle: NSTextField!
@@ -25,7 +31,8 @@ class PullRequestCell: NSTableCellView {
 }
 
 class ViewController: NSViewController, NSTableViewDataSource, NSTableViewDelegate {
-    @IBOutlet weak var tableViewContainer: NSScrollView!
+    @IBOutlet weak var branchSelector: NSPopUpButton!
+    @IBOutlet weak var tableViewContainer: NSView!
     @IBOutlet weak var tableView: NSTableView!
     @IBOutlet weak var backgroundImage: NSImageView!
     @IBOutlet weak var backgroundLabel: NSTextField!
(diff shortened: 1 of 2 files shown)


WallEView/ViewController.swift
```diff
@@ -111,20 +121,28 @@ class ViewController: NSViewController, NSTableViewDataSource, NSTableViewDelega
...
+                        self.branchSelector.menu?.items = queues.map {
+                            NSMenuItem(title: $0.targetBranch, action: #selector(self.switchBranch(_:)), keyEquivalent: "")
+                        }
                     } catch {
-                        self?.state = .empty
+                        self.state = State(queues: [], error: error, targetBranch: nil)
                     }
                 } else {
-                    self?.state = .empty
+                    self.state = State(queues: [], error: error, targetBranch: nil)
```
AliSoftware, line 137: I tried disconnecting the VPN then opening the WallEView MenuItem but all I got was an empty view, no error shown on the UI.

After quick testing it seems that with VPN off, what we get is a "Request Timeout" on the `URLSessionDataTask` since this is what I get in the console:

```
2019-12-03 14:20:04.130874+0100 WallEView[15038:198736] Task <BFCB74DF-7E9B-46E8-81F0-83E297000551>.<1> finished with error [-1001] Error Domain=NSURLErrorDomain Code=-1001 "The request timed out." UserInfo={NSUnderlyingError=0x600000ccd6b0 {Error Domain=kCFErrorDomainCFNetwork Code=-1001 "(null)" UserInfo={_kCFStreamErrorCodeKey=-2102, _kCFStreamErrorDomainKey=4}}, NSErrorFailingURLStringKey=https://ios-merge-bot.ops.babylontech.co.uk/, NSErrorFailingURLKey=https://ios-merge-bot.ops.babylontech.co.uk/, _kCFStreamErrorDomainKey=4, _kCFStreamErrorCodeKey=-2102, NSLocalizedDescription=The request timed out.}
2019-12-03 14:20:04.132399+0100 WallEView[15038:198742] Task <BFCB74DF-7E9B-46E8-81F0-83E297000551>.<1> HTTP load failed, 0/0 bytes (error code: -999 [1:89])
```

But by the time we get to this error `self` is back to `nil` so `guard let self = self` fails and exits early (yes, even if the MenuItem was still open the whole time and never closed)

AliSoftware (commented):
Nice! Just hit an issue with error reporting wrt VPN though

WallEView/ViewController.swift
ilyapuchka: is it when you dismiss the popup? if you keep the popup self shouldn't be deallocated, otherwise when it is to be displayed again it is recreated (changing this would be out of scope)

AliSoftware: As said above:

>  (yes, even if the MenuItem was still open the whole time and never closed)

Which was what surprised me the most indeed

ilyapuchka: ok, good to know, I'll take a look at that later, I think we can leave without that for now.

ilyapuchka: I've just tried to replicate this with httpbin and error is reported properly in UI 🤷‍♂

AliSoftware: Ok I just retried too and ended up seeing the error too.

I think maybe it was because I had a breakpoint, so maybe something along the line of: the ViewController had time to deallocate just right when the focus changed to Xcode when it hit the breakpoint, but the popup was still on screen because the runloop was interrupted by the debugger so the view itself was still visible on screen?

In any case, I can see the error reported in the UI now when I disable all my breakpoints.

Would still be nice maybe to reduce the timeout to 20s or something (rather than the default 60s) because many of us are on the VPN and that's an error that is likely to happen often (people forgetting to turn on VPN) and having to wait 60s with the popup open just to understand what's wrong is not ideal… (And maybe later use some Reachability model to detect faster when host is unreachable? 🤔)

ilyapuchka: I had a similar issue with breakpoints.
I think all this are good ideas, but they are out of the scope of this PR that fixes the issue with decoding branch property and adds UI to switch between queues that we just introduced. We can improve other aspects of experience later.

AliSoftware (approved):

WallEView/ViewController.swift
```diff
@@ -35,31 +42,36 @@ class ViewController: NSViewController, NSTableViewDataSource, NSTableViewDelega
...
             tableViewContainer.isHidden = hideContent
 
-            if state?.queues.first?.isFailing == true {
+            if state.isFailing == true {
                 backgroundImage.image = #imageLiteral(resourceName: "foot")
-                backgroundLabel.stringValue = "Something failin':\n\n\(queue?.error.map(String.init(describing:)) ?? "")"
+                backgroundLabel.stringValue = "Something failin':\n\n\(state.error.map(String.init(describing:)) ?? "")"
             } else {
                 backgroundImage.image = #imageLiteral(resourceName: "green")
                 backgroundLabel.stringValue = "Doin' nothin', just chillin'"
```
NSMyself, line 58: 🍸

NSMyself (approved):
:godmode: good job man

PR #45, opened 2019-12-03
[IOSP-437] Fix logic when creating MergeService (leading to auto-accepting PRs for new target branches)

AliSoftware:
Ticket: https://babylonpartners.atlassian.net/browse/IOSP-437

After deploying #39 recently, we realised that there was a flaw in the logic: every time a new Pull Request targeting a branch that doesn't exist yet is created, leading to the creation on a new MergeService, we forgot to filter out those PullRequest events sent by GitHub to only create the new MergeService if that PR was supposed to be included in the first place)

Consequently, any new PR targeting a new target branch was automatically triggering the creation of a new `MergeService` and add that PR to the `MergeService`'s queue… even if that PR didn't have the integration label and wasn't ready to be merged (basically as soon as the PR is created)

This PR fixes that flawed logic.

### DIFF
--- a/Sources/Bot/Services/DispatchService.swift
+++ b/Sources/Bot/Services/DispatchService.swift
@@ -88,7 +88,7 @@ public final class DispatchService {
                 return service
             } else {
                 // Only create a new MergeService is the PullRequest for which we received an event is supposed to be included
-                guard case .include(let pullRequest) = MergeService.Event.Outcome(event: event, integrationLabel: self.integrationLabel) else {
+                guard case .include(let pullRequest)? = MergeService.Event.Outcome(event: event, integrationLabel: self.integrationLabel) else {
                     return nil
                 }
 
--- a/Sources/Bot/Services/DispatchService.swift
+++ b/Sources/Bot/Services/DispatchService.swift
@@ -87,7 +87,7 @@ public final class DispatchService {
                 // If service was already existing, return it so we'll send the pullRequestChangesObserver event outside this `modify` below
                 return service
             } else {
-                // Only create a new MergeService is the PullRequest for which we received an event is supposed to be included
+                // Only create a new MergeService if the PullRequest for which we received an event is supposed to be included
                 guard case .include(let pullRequest)? = MergeService.Event.Outcome(event: event, integrationLabel: self.integrationLabel) else {
                     return nil
                 }
--- a/Sources/Bot/Services/DispatchService.swift
+++ b/Sources/Bot/Services/DispatchService.swift
@@ -80,16 +80,22 @@ public final class DispatchService {
 
     private func pullRequestDidChange(event: PullRequestEvent) {
         logger.log("📣 Pull Request did change \(event.pullRequestMetadata) with action `\(event.action)`")
+
         let targetBranch = event.pullRequestMetadata.reference.target.ref
         let existingService = mergeServices.modify { (dict: inout [String: MergeService]) -> MergeService? in
             if let service = dict[targetBranch] {
                 // If service was already existing, return it so we'll send the pullRequestChangesObserver event outside this `modify` below
                 return service
             } else {
+                // Only create a new MergeService is the PullRequest for which we received an event is supposed to be included
+                guard case .include(let pullRequest) = MergeService.Event.Outcome(event: event, integrationLabel: self.integrationLabel) else {
+                    return nil
+                }
+
                 dict[targetBranch] = makeMergeService(
                     targetBranch: targetBranch,
                     scheduler: self.scheduler,
-                    initialPullRequests: [event.pullRequestMetadata.reference]
+                    initialPullRequests: [pullRequest]
                 )
                 // If MergeService didn't exist yet and we just created it, return nil so that we DON'T send the event on pullRequestChangesObserver
                 // outside this `modify` below (because we already passed the PR to initialPullRequests parameters when creating the service – and
@@ -111,6 +117,7 @@ public final class DispatchService {
     }
 
     private func makeMergeService(targetBranch: String, scheduler: DateScheduler, initialPullRequests: [PullRequest] = []) -> MergeService {
+        logger.log("🆕 New MergeService created for target branch `\(targetBranch)`")
         let mergeService = MergeService(
             targetBranch: targetBranch,
             integrationLabel: integrationLabel,
@@ -125,9 +132,10 @@ public final class DispatchService {
         mergeServiceLifecycleObserver.send(value: .created(mergeService))
         mergeService.state.producer
             .observe(on: scheduler)
-            .startWithValues { [weak self, service = mergeService] state in
+            .startWithValues { [weak self, service = mergeService, logger = logger] state in
                 self?.mergeServiceLifecycleObserver.send(value: .stateChanged(service))
                 if state.status == .idle {
+                    logger.log("👋 MergeService for target branch `\(targetBranch)` is idle, destroying")
                     self?.mergeServices.modify { dict in
                         dict[targetBranch] = nil
                     }
--- a/Sources/Bot/Services/DispatchService.swift
+++ b/Sources/Bot/Services/DispatchService.swift
@@ -43,9 +43,6 @@ public final class DispatchService {
 
         gitHubAPI.fetchPullRequests()
             .flatMapError { _ in .value([]) }
-            .map { pullRequests in
-                pullRequests.filter { $0.isLabelled(with: self.integrationLabel) }
-            }
             .observe(on: scheduler)
             .startWithValues { pullRequests in
                 self.dispatchInitial(pullRequests: pullRequests)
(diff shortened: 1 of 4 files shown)


Sources/Bot/Services/MergeService.swift
```diff
@@ -279,6 +279,27 @@ extension MergeService {
     }
 }
 
+extension MergeService.Event.Outcome {
```
dmcrodrigues: Do we really need a type for this? Why not using free/static functions? I don't find it a type particularly discoverable 😕

dmcrodrigues: When I say a type I mean using failable initialisers.

AliSoftware: tbh I initially created a `static` free function at first, but the signature felt a bit strange to read, especially since I named that `static func outcome(for: …) -> MergeService.Event.Outcome?`… so since the func was named `outcome` and returned the (already existing †) `Outcome` type, I figured it would make more sense to make it an initialiser on the existing type

_(†) The `Outcome` type is not new and used in the `Feedbacks` of the `MergeService` machine already, all I did is move the logic from the `Feedback` to a `init?` to be able to _also_ call it from the `DispatchService` to decide wether to create a new queue or not)_

I'm ok with moving those to free functions if you prefer it to using an `init?` – just know that even if I do that the `MergeService.Event.Outcome` type will still exist as it's used in the state machine already.

Sources/Bot/Services/DispatchService.swift
```diff
@@ -80,16 +80,22 @@ public final class DispatchService {
...
     private func pullRequestDidChange(event: PullRequestEvent) {
         logger.log("📣 Pull Request did change \(event.pullRequestMetadata) with action `\(event.action)`")
+
         let targetBranch = event.pullRequestMetadata.reference.target.ref
         let existingService = mergeServices.modify { (dict: inout [String: MergeService]) -> MergeService? in
             if let service = dict[targetBranch] {
                 // If service was already existing, return it so we'll send the pullRequestChangesObserver event outside this `modify` below
                 return service
             } else {
+                // Only create a new MergeService is the PullRequest for which we received an event is supposed to be included
```
AliSoftware:
```suggestion
                // Only create a new MergeService if the PullRequest for which we received an event is supposed to be included
```

Sources/Bot/Services/MergeService.swift
```diff
@@ -565,19 +586,8 @@ extension MergeService {
...
-                    case .labeled where metadata.reference.isLabelled(with: state.integrationLabel) && metadata.isMerged == false:
-                        return Event.Outcome.include(metadata.reference)
-                    case .unlabeled where metadata.reference.isLabelled(with: state.integrationLabel) == false:
-                        return Event.Outcome.exclude(metadata.reference)
-                    case .closed:
-                        return Event.Outcome.exclude(metadata.reference)
-                    default:
-                        return nil
-                    }
+                .map { metadata, action in
```
dmcrodrigues: `map`+`skipNil` can be flatten using `filterMap`.

Sources/Bot/Services/MergeService.swift
dmcrodrigues: That's ok let's keep it, is just a detail 👍

Sources/Bot/Services/MergeService.swift
AliSoftware: Good point, not sure why we didn't spot it earlier 👍

Sources/Bot/Services/DispatchService.swift
```diff
@@ -80,16 +80,22 @@ public final class DispatchService {
...
         logger.log("📣 Pull Request did change \(event.pullRequestMetadata) with action `\(event.action)`")
+
         let targetBranch = event.pullRequestMetadata.reference.target.ref
         let existingService = mergeServices.modify { (dict: inout [String: MergeService]) -> MergeService? in
             if let service = dict[targetBranch] {
                 // If service was already existing, return it so we'll send the pullRequestChangesObserver event outside this `modify` below
                 return service
             } else {
+                // Only create a new MergeService if the PullRequest for which we received an event is supposed to be included
+                guard case .include(let pullRequest)? = MergeService.Event.Outcome(event: event, integrationLabel: self.integrationLabel) else {
```
dmcrodrigues: Aren't we leaking implementation details in doing this filtering at this point? Because `MergeService.Event.Outcome` should be something private to `MergeService` and not exposed even internally, same for Events which are now exposed. 

I understand that we may end up creating a new `MergeService` without a need for it, if the PR is not labelled with the integration label, but that's an implementation detail and the `MergeService` will be released eventually so that's not entirely harmful in creating this instance.

AliSoftware: So you're suggesting we create the MergeService every time a new PR targeting a new branch is created, even if that merge service will directly go from `.starting` to `.idle` and be deleted right away?

I'm ok with that, just wondering if it's worth it compared to refactor that to just expose a `MergeService.shouldInclude(…) -> Bool` func (that would privately check if the `Outcome` is `.included` but will keep `Outcome` private?

dmcrodrigues: Either we do it for all events or none in my opinion, as a reader/maintainer is very confusing why we only filter part of the events... Because if the service already exists we just forward the change to the service which will eventually filter the event internally in the feedback that I've mentioned.

```swift
existingService?.pullRequestChangesObserver.send(value: (event.pullRequestMetadata, event.action))
```

Conceptually `DispatchService` should only be responsible by keeping the merge services and forward the events accordingly, now it's already doing some logic where it becomes aware of the concept of filtering Pull Requests that are not labelled with a certain integration label which is a concept that's expected to be part of the service itself.

AliSoftware: Yes I agree that I'm not a fan of that logic of only forwarding the event to the service if it already existed and use the `initialPullRequests` otherwise. We might want to improve that indeed.

I'll try to work on this, allowing the `.starting` state to also be able to listen and react to `pullRequestEvents`. _(Tried that in the original implementation but let to some unexpected complications about tracking state changes during startup sequence 😉 will try to revisit now that the code has settled)_

dmcrodrigues: > I'll try to work on this, allowing the .starting state to also be able to listen and react to pullRequestEvents.

I'm not entirely following, we are injecting `initialPullRequests` to the `MergeService` as part of the initialisation so that filtering can be done directly in the initialiser or inside the feedback for `starting` status, I'm not sure why we need to listen and react to `pullRequestEvents` 😕

AliSoftware: ~We can't really do the filtering about the type of PullRequest event in `initialPullRequests` since we'd have the PullRequests directly here, not the event, so we wouldn't be able to check if the event leading to this was a `.opened` or `.labeled` or whatnot~

Wait or were you suggesting to just filter on the `pullRequest.isLabelled(state.integrationlabel)` – instead of filtering on the event? (I thought that at first but was afraid of possible edge cases, but in fact maybe that could be enough indeed 🤔 )

dmcrodrigues: Events are irrelevant for the initial list of pull requests, the only thing that matters is if the PR has the integration label or not, during the start up of the bot we don't have any events being fired, we just load the entire list of pull requests and start from there based on the presence of the integration label.

AliSoftware: Gotcha. But that means if we go that route, we'll keep the `existingService?.pullRequestChangesObserver.send` iow the fact that the `DispatchService` knows when to send those events (when MergeService was already existing) vs when not to (when it just got created), and I didn't like that either and wanted to improve this. But maybe I can send the event in all cases and filter in the `Feedback` if it's `.starting` after all, so this should work.

Sources/Bot/Services/DispatchService.swift
```diff
@@ -80,29 +77,25 @@ public final class DispatchService {
...
-                // If MergeService didn't exist yet and we just created it, return nil so that we DON'T send the event on pullRequestChangesObserver
-                // outside this `modify` below (because we already passed the PR to initialPullRequests parameters when creating the service – and
-                // the service would still be `.starting` and it would not be ready to receive those events anyway)
-                return nil
+                dict[targetBranch] = newService
+                return newService
             }
         }
-        existingService?.pullRequestChangesObserver.send(value: (event.pullRequestMetadata, event.action))
+        existingService.pullRequestChangesObserver.send(value: (event.pullRequestMetadata, event.action))
```
AliSoftware, line 95: Notice that we now send the event in all cases, and let the `MergeService` filter that event out (the `Feedback.pullRequestChanges` of `MergeService` already filters out those events if the service was just started, using `predicate: { $0.status != .starting }`).

Which means we don't leak logic and implementation details about the internals of `MergeService` to `DispatchService` anymore, it doesn't have to know that logic about "don't send event if service was just created", instead we dispatch in all cases and let the `MergeService` do the filtering

Sources/Bot/Services/MergeService.swift
```diff
@@ -43,12 +43,14 @@ public final class MergeService {
             statusChecksTimeout: statusChecksTimeout
         )
 
+        let pullRequestsReadyToInclude = initialPullRequests.filter { $0.isLabelled(with: integrationLabel) }
```
AliSoftware, line 46: This ends up being the main fix for the initial bug

Tests/BotTests/MergeService/DispatchServiceTests.swift
```diff
@@ -140,6 +140,88 @@ class DispatchServiceTests: XCTestCase {
...
+                    .state(.stub(targetBranch: developBranch, status: .runningStatusChecks(dev1.with(mergeState: .blocked)))),
+                    .state(.stub(targetBranch: developBranch, status: .runningStatusChecks(dev1.with(mergeState: .blocked)), pullRequests: [dev2.reference])),
+
+                    // MergeService is always created when we receive a new PR event…
+                    .created(branch: releaseBranch),
+                    .state(.stub(targetBranch: releaseBranch, status: .starting)),
+                    // … but since that new PR got filtered out for not having the integration label…
+                    .state(.stub(targetBranch: releaseBranch, status: .idle)),
+                    // … then the service is destroyed right away.
+                    .destroyed(branch: releaseBranch),
```
AliSoftware, line 212: Notice the new behavior: with the new implementation that avoids `DispatchService` having to know about the internal logic and impl details of `MergeService`, that means that it always creates a `MergeService` even if that `MergeService` ends up deciding there's nothing to do and gets immediately destroyed 🤷‍♂

Tests/BotTests/MergeService/MergeServiceTests.swift
```diff
@@ -239,6 +244,12 @@ class MergeServiceTests: XCTestCase {
...
             assert: {
                 expect($0) == [
+                    // First received new PR with no integration label, so we end up creating the MergeService but destroying it right away since the queue is empty
+                    .created(branch: MergeServiceFixture.defaultTargetBranch),
+                    .state(.stub(status: .starting)),
+                    .state(.stub(status: .idle)),
+                    .destroyed(branch: MergeServiceFixture.defaultTargetBranch),
+                    // Then received PR event about adding integration label, so we end up creating the MergeService again but this time for good
                     .created(branch: MergeServiceFixture.defaultTargetBranch),
                     .state(.stub(status: .starting)),
```
AliSoftware, line 254: Notice the new behavior: with the new implementation that avoids `DispatchService` having to know about the internal logic and impl details of `MergeService`, that means that it always creates a `MergeService` even if that `MergeService` ends up deciding there's nothing to do and gets immediately destroyed… and gets recreated soon after when we receive the `.labeled` event.

AliSoftware (commented):
@dmcrodrigues New commits should address your feedback 👍

Sources/Bot/Services/DispatchService.swift
```diff
@@ -43,9 +43,6 @@ public final class DispatchService {
 
         gitHubAPI.fetchPullRequests()
             .flatMapError { _ in .value([]) }
-            .map { pullRequests in
-                pullRequests.filter { $0.isLabelled(with: self.integrationLabel) }
-            }
```
AliSoftware, line 48: `DispatchService` shouldn't know about the logic related to deciding which PR to filter in or out, which is why this has been removed.
That filtering will be done by the `MergeService` directly instead, to avoid leaking any implementation details and logic of the `MergeService` to the `DispatchService` (even if that means creating a `MergeService` only to see it immediately go to its `.idle` state and being destroyed)

Tests/BotTests/MergeService/DispatchServiceTests.swift
dmcrodrigues: For now... we shouldn't exclude doing something in the future, imagine if we introduce the ability to monitor the PRs to check the number of reviews pending and notify that the PR is ready to be integrated? e.g. after 2 approvals In that scenario the service may stay alive.

dmcrodrigues (approved):

danspndl (approved):
LGTM 🚀

ilyapuchka (approved):

inamiy (approved):

PR #47, opened 2019-12-05
[IOSP-443] fix tests on linux

AliSoftware:
`DispatchServiceTests` were not added to the XCTestManifest so not run on Linux.

It appears that despite them passing consistently on macOS, they only fail on Linux.
This PR attempts to debug and fix that.


dmcrodrigues:
@AliSoftware are you still working on this one?

AliSoftware:
Ok I think I start to understand what might be going on here, but not sure how to solve it in a proper manner. If my investigation is right, the issue is mainly that when using a `TestScheduler`, some Signals used in our state machines emit multiple values at the _exact same time_.

* e.g. when we start the bot with 3 PRs for 3 different branches all ready to be merged, then all the `MergeServices` call the `MockGitHubAPI` on the same scheduler's tick. This might mess up with our `MockGitHubAPI` which expect a deterministic order on which the APIs are getting called
* In each State Machine, if we receive 2 events at the exact same time on our event Signal, it becomes non-deterministic which of the 2 simultaneous event is processed first: on Mac it's one, on Linux it can be the other, leading to a different state in the state machine

Statistically, such a use case probably can't really happen with a real scheduler and GitHub API network requests that will never take the exact same time down to the smallest time unit possible. But that's still something we should potentially account for in RAF code.

This theory is still to be proven; I'm planning to do some more tests to check it first

AliSoftware:
Ticket returned to backlog as I spent too much time and experiments on it without managing to fix it (and getting more and more obscure behaviors when trying). Will keep the branch open so that we have a history of what was tried though.

Sources/Bot/Services/MergeService.swift
```diff
@@ -390,8 +394,11 @@ extension MergeService {
 
         return Feedback(skippingRepeated: { $0.status.integrationMetadata }) { metadata -> SignalProducer<Event, Never> in
 
-            guard metadata.isMerged == false
-                else { return .value(.integrationDidChangeStatus(.done, metadata)) }
+            guard metadata.isMerged == false else {
+                return SignalProducer
```
dmcrodrigues, line 398: The compiler was complaining about this?

Sources/Bot/Services/MergeService.swift
```diff
@@ -543,7 +548,7 @@ extension MergeService {
                             return requiredStateProducer.zip(with: .value(pullRequestMetadataRefreshed))
                         }
                         .flatMapError { _ in .empty }
-                        .filterMap { state, pullRequestMetadataRefreshed in
+                        .compactMap { state, pullRequestMetadataRefreshed in
```
dmcrodrigues, line 551: Good stuff, I forgot to change this.

dmcrodrigues: @AliSoftware there's an additional `filterMap` that needs to be changed IIRC can you check and change it too?

Sources/Bot/Services/MergeService.swift
AliSoftware: It's because I added a missing `observe(on:)` to the chain. So now we can't use the shorthand dot-leading syntax since it's a chain and not a single expression

Sources/Bot/Services/MergeService.swift
dmcrodrigues: Just a warning but still.

AliSoftware: Good catch. f9ad7a4208d2c2cdb29ca12f7bcbe0cb908dfd3f

dmcrodrigues (approved):

Tests/BotTests/XCTestManifests.swift
```diff
@@ -1,6 +1,21 @@
 #if !canImport(ObjectiveC)
 import XCTest
 
+extension DispatchServiceTests {
+    // DO NOT MODIFY: This is autogenerated, use:
+    //   `swift test --generate-linuxmain`
+    // to regenerate.
+    static let __allTests__DispatchServiceTests = [
```
ilyapuchka, line 8: btw didn't they release automatic test discovery on linux? or it's just in master?

ilyapuchka (approved):

Tests/BotTests/XCTestManifests.swift
AliSoftware: Was not aware of that, if that's the case that would be wonderful, given how many times we forgot to update those files!

ilyapuchka: some info here https://forums.swift.org/t/test-discovery-on-linux/26203 it's actually SPM feature

AliSoftware: Cool, thanks for the link; that looks promising!
Not sure what to conclude from that thread regarding if it's already released (or still only on master though), but probably not yet; but clearly something to keep a look at 👍

PR #48, opened 2019-12-13
[IOSP-467] [IOSP-468] Prevent MergeBot to comment on closed PRs after merge

AliSoftware:
Tickets:
 - https://babylonpartners.atlassian.net/browse/IOSP-467
 - https://babylonpartners.atlassian.net/browse/IOSP-468

## Why?

[IOSP-467] Since recently the bot have been commenting on PRs immediately after them being merged, telling `"Your pull request was accepted and is going to be handled right away 🏎"`… as if the PR was re-added to the queue immediately after being merged

This was because GitHub events for PRs being closed were not filtered out when forwarded to a newly-created MergeService.

## How

* We now forward GitHub `PullRequestEvents` _even_ right after having started a new `MergeService`, so that the GH event goes through the normal state machine – and via the `Feedback.pullRequestChanges` – even immediately after the `MergeService` creation (as opposed to the previous implementation which made a special case of the initialisation and initial Pull Requests passed during a `MergeService`'s instantiation by the `DispatchService` on new event)
* The `MergeService` still processes the initial Pull Requests sent by the `DispatchService` BUT those are *only* sent during the bot's boot sequence – because those don't come from any GitHub webhook/event, but come from a call to the API (to get all open Pull Requests) at bot startup.

---

[IOSP-468] We also now wait for a configurable delay (defaults to 5 minutes) before killing a idle MergeService, to avoid constantly killing and recreating them when there is still some activity on a given branch and queue (like GitHub events close to one another)

---

### Quick and Dirty Fix

Initial quick-fix for IOSP-467 was to just filter-out `.closed` events in the `DispatchService` to avoid creating and dispatching the event to the `MergeService` and thus avoid the comment.

But that meant that we'd leak some logic and implementation details to the `DispatchService`, while it should be up to the `MergeService` to decide if it wants to do something with closed events. Besides, any other event (like `.unlabelled`) could also trigger that bug, not just `.closed` – even though that's the main case where we saw it happen in practice.

### Getting rid of the `.starting` state

I also tried to get rid of the `.starting` state altogether and directly start at the `.idle` or `.ready` state, since this `.starting` state is now only used for the bot's boot sequence (when the list of all open PRs is fetched for initial re-sync)

But the issue is we need a state transition to happen (between `.starting` and `.idle`/`.ready`) at boot, in order to trigger the `Feedback.whenAddingPullRequests` (which derives from `combinePrevious()`) *even during that boot sequence*. Otherwise the service won't post the "Your PR was accepted…" comments on the PRs that are added right after a bot's reboot

### Always starting a new `MergeService` in `.starting` state

If we decide instead that, since we are going to keep the `.starting` state, we might as well always start with it even when the `MergeService` was instantiated after a GitHub event (and not just during the bot's reboot), and let the `MergeService`'s state machine move us to the right state on the next scheduler's tick… then we end up with *two* `Events` emitted by the `Feedbacks` at the same time during the `.starting` state:
 - `.pullRequestsLoaded([])` returned from `whenStarting`
 - `.pullRequestDidChange` return from `pullRequestChanges` (because `DispatchService` forwards the event after having created the `MergeService` if it didn't exist)

Those are emitted exactly at the same time and `.merge()`'d by RAF, and will end up being processed in an _arbitrary_ order – both racing each other, or even one of them ending up being ignored by the state machine. Besides, this is making the state machine non-deterministic (i.e. the next state would depend on which event gets processed first – which is different on Mac vs Linux btw).

This is why this solution has been rejected and we start with `.idle` instead of `.starting` when the `MergeService` is created outside of the bot's reboot sequence, as the only path transitioning out of `.idle` is the `pullRequestDidChange` event and we don't risk to also react to `pullRequestsLoaded` (used to transition out of `.starting`) in that use case.

[IOSP-467]: https://babylonpartners.atlassian.net/browse/IOSP-467
[IOSP-468]: https://babylonpartners.atlassian.net/browse/IOSP-468

### DIFF
--- a/Sources/App/Extensions/EnvironmentProperties.swift
+++ b/Sources/App/Extensions/EnvironmentProperties.swift
@@ -31,6 +31,12 @@ extension Environment {
         return TimeInterval(value)
     }
 
+    /// Delay to wait after a MergeService is back in idle state before destroying it
+    static func idleMergeServiceCleanupDelay() throws -> TimeInterval? {
+        let value: String = try Environment.get("IDLE_BRANCH_QUEUE_CLEANUP_DELAY")
+        return TimeInterval(value)
+    }
+
     static func mergeLabel() throws -> PullRequest.Label {
         return PullRequest.Label(name: try Environment.get("MERGE_LABEL"))
     }
--- a/Sources/Bot/Services/DispatchService.swift
+++ b/Sources/Bot/Services/DispatchService.swift
@@ -77,6 +77,7 @@ public final class DispatchService {
 
     private func pullRequestDidChange(event: PullRequestEvent) {
         logger.log("📣 Pull Request did change \(event.pullRequestMetadata) with action `\(event.action)`")
+        guard event.action != .closed else { return }
 
         let targetBranch = event.pullRequestMetadata.reference.target.ref
         let existingService = mergeServices.modify { (dict: inout [String: MergeService]) -> MergeService in
--- a/Sources/Bot/Services/DispatchService.swift
+++ b/Sources/Bot/Services/DispatchService.swift
@@ -133,8 +133,8 @@ public final class DispatchService {
             }
         // Observe idle states to clean up dormant MergeServices only after they have been idle for too long
         mergeService.state.producer
-            .filter { $0.status == .idle }
             .debounce(self.idleMergeServiceCleanupDelay, on: scheduler)
+            .filter { $0.status == .idle }
             .startWithValues { [weak self, service = mergeService, logger = logger] state in
                 guard let self = self else { return }
 
--- a/Sources/Bot/Services/DispatchService.swift
+++ b/Sources/Bot/Services/DispatchService.swift
@@ -69,7 +69,7 @@ public final class DispatchService {
                 dict[branch] = makeMergeService(
                     targetBranch: branch,
                     scheduler: self.scheduler,
-                    initialTrigger: .booting(openPullRequests: pullRequestsForBranch)
+                    initialPullRequests: pullRequestsForBranch
                 )
             }
         }
@@ -85,8 +85,7 @@ public final class DispatchService {
             } else {
                 let newService = makeMergeService(
                     targetBranch: targetBranch,
-                    scheduler: self.scheduler,
-                    initialTrigger: .gitHubEvent(event)
+                    scheduler: self.scheduler
                 )
                 dict[targetBranch] = newService
                 return newService
@@ -106,7 +105,7 @@ public final class DispatchService {
     private func makeMergeService(
         targetBranch: String,
         scheduler: DateScheduler,
-        initialTrigger: MergeService.InitialTrigger
+        initialPullRequests: [PullRequest] = []
     ) -> MergeService {
         logger.log("🆕 New MergeService created for target branch `\(targetBranch)`")
         let mergeService = MergeService(
@@ -115,7 +114,7 @@ public final class DispatchService {
             topPriorityLabels: topPriorityLabels,
             requiresAllStatusChecks: requiresAllStatusChecks,
             statusChecksTimeout: statusChecksTimeout,
-            initialTrigger: initialTrigger,
+            initialPullRequests: initialPullRequests,
             logger: logger,
             gitHubAPI: gitHubAPI,
             scheduler: scheduler
--- a/Sources/Bot/Services/DispatchService.swift
+++ b/Sources/Bot/Services/DispatchService.swift
@@ -134,7 +134,6 @@ public final class DispatchService {
         // Observe idle states to clean up dormant MergeServices only after they have been idle for too long
         mergeService.state.producer
             .filter { $0.status == .idle }
-            .skipRepeats()
             .debounce(self.idleMergeServiceCleanupDelay, on: scheduler)
             .startWithValues { [weak self, service = mergeService, logger = logger] state in
                 guard let self = self else { return }
--- a/Sources/Bot/Services/DispatchService.swift
+++ b/Sources/Bot/Services/DispatchService.swift
@@ -69,7 +69,7 @@ public final class DispatchService {
                 dict[branch] = makeMergeService(
                     targetBranch: branch,
                     scheduler: self.scheduler,
-                    initialPullRequests: pullRequestsForBranch
+                    initialTrigger: .booting(openPullRequests: pullRequestsForBranch)
                 )
             }
         }
@@ -77,7 +77,6 @@ public final class DispatchService {
 
     private func pullRequestDidChange(event: PullRequestEvent) {
         logger.log("📣 Pull Request did change \(event.pullRequestMetadata) with action `\(event.action)`")
-        guard event.action != .closed else { return }
 
         let targetBranch = event.pullRequestMetadata.reference.target.ref
         let existingService = mergeServices.modify { (dict: inout [String: MergeService]) -> MergeService in
@@ -87,7 +86,7 @@ public final class DispatchService {
                 let newService = makeMergeService(
                     targetBranch: targetBranch,
                     scheduler: self.scheduler,
-                    initialPullRequests: [event.pullRequestMetadata.reference]
+                    initialTrigger: .gitHubEvent(event)
                 )
                 dict[targetBranch] = newService
                 return newService
@@ -104,7 +103,11 @@ public final class DispatchService {
         }
     }
 
-    private func makeMergeService(targetBranch: String, scheduler: DateScheduler, initialPullRequests: [PullRequest] = []) -> MergeService {
+    private func makeMergeService(
+        targetBranch: String,
+        scheduler: DateScheduler,
+        initialTrigger: MergeService.InitialTrigger
+    ) -> MergeService {
         logger.log("🆕 New MergeService created for target branch `\(targetBranch)`")
         let mergeService = MergeService(
             targetBranch: targetBranch,
@@ -112,7 +115,7 @@ public final class DispatchService {
             topPriorityLabels: topPriorityLabels,
             requiresAllStatusChecks: requiresAllStatusChecks,
             statusChecksTimeout: statusChecksTimeout,
-            initialPullRequests: initialPullRequests,
+            initialTrigger: initialTrigger,
             logger: logger,
             gitHubAPI: gitHubAPI,
             scheduler: scheduler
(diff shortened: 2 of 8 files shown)


dmcrodrigues:
@AliSoftware can we have a test? Could be just creating a `MergeService` with a PR already merged.

AliSoftware:
Actually just detected another bug, should not just filter out closed events but also unlabeled and more.
Will fix and add test indeed.

Sources/Bot/Services/DispatchService.swift
```diff
@@ -85,8 +88,7 @@ public final class DispatchService {
             } else {
                 let newService = makeMergeService(
                     targetBranch: targetBranch,
-                    scheduler: self.scheduler,
-                    initialPullRequests: [event.pullRequestMetadata.reference]
+                    scheduler: self.scheduler
```
AliSoftware, line 91: ℹ️ Note: the event is now just forwarded (see `mergeService.pullRequestChangesObserver.send(…)` below) in *all* cases (existing OR newly-created service) and is processed by the `MergeService` in both cases, which means we don't need to pass it to `initialPullRequests` anymore and make a special case. Instead this is now consistent as we _always_ just forward the GH event (wether the MergeService already existed for that branch or we spawn a new one)

By contrast, in the previous implementation:
 * the event was ignored when the `MergeService` was in `.starting` state which is why we passed it in `initialPullRequests` in the case a MS didn't exist yet and we spawned a new one…
 * but that former logic was not consistent and made a special case out of this, leading to the inconsistencies in how PRs were treated by existing vs newly-spawed `MergeServices`…  (going thru logic in `init` vs going via a `Feedback`, respectively) which is what led to that IOSP-467 bug in the end

Sources/App/Extensions/EnvironmentProperties.swift
```diff
@@ -31,6 +31,12 @@ extension Environment {
         return TimeInterval(value)
     }
 
+    /// Delay to wait after a MergeService is back in idle state before destroying it
+    static func idleMergeServiceCleanupDelay() throws -> TimeInterval? {
```
ilyapuchka, line 35: [nit] I'd name this thing something like "keep alive timeout"

Sources/Bot/Services/DispatchService.swift
```diff
@@ -116,18 +122,27 @@ public final class DispatchService {
             gitHubAPI: gitHubAPI,
             scheduler: scheduler
         )
+
+        // Forward events about creation and subsequent state changes of the new MS to our lifecycleObserver
```
ilyapuchka: [nit]
```suggestion
        // Forward events about creation and subsequent state changes of the new merge service to our lifecycleObserver
```

Sources/Bot/Services/MergeService.swift
```diff
@@ -17,13 +17,33 @@ public final class MergeService {
...
+    ///
+    /// A MergeService is a state machine responsible for handling a queue of Pull Requests,
+    /// merging them in the right order to avoid multiple PRs from conflicting or affecting one another once merged.
+    ///
+    /// An instance of a MergeService is responsible for all the Pull Requests targeting the branch it is responsible for.
+    /// There can be multiple MergeServices instances in parallel, each handling the PRs for a different target branch,
+    /// (since it's still safe to merge Pull Requests targeting different branches in parallel).
+    /// Those different MergeServices per target branch are typically managed by a `DispatchService`
+    ///
+    /// - Parameter targetBranch: The target branch this MergeService is responsible for
```
ilyapuchka: [nit] there is a format to group all parameters without prefixing each line (might be a bit different that the suggestion in reality)
 
```suggestion
    /// Parameters:
    /// - targetBranch: The target branch this MergeService is responsible for
```

Sources/Bot/Services/MergeService.swift
```diff
@@ -132,14 +153,20 @@ extension MergeService {
             self.status = status
         }
 
-        static func initial(targetBranch: String, integrationLabel: PullRequest.Label, topPriorityLabels: [PullRequest.Label], statusChecksTimeout: TimeInterval) -> State {
+        static func initial(
+            targetBranch: String,
+            integrationLabel: PullRequest.Label,
+            topPriorityLabels: [PullRequest.Label],
+            statusChecksTimeout: TimeInterval,
+            isWaitingForInitialPullRequests: Bool
```
ilyapuchka: [nit] I fill like naming this parameter as `isStarting` or `isIdle` will make it more clear

Tests/BotTests/MergeService/DispatchServiceContext.swift
```diff
@@ -51,4 +53,9 @@ class DispatchServiceContext {
                 self?.events.append(event)
             }
     }
+
+    static func idleCleanupDelay(times: Double = 1.0) -> DispatchTimeInterval {
```
ilyapuchka: I'm not quite sure why we need to multiply the timeout?

ilyapuchka (approved):
I'm not sure why we need to additionally multiply the time interval instead of just increasing it but otherwise looks ok

Sources/App/Extensions/EnvironmentProperties.swift
AliSoftware: I thought about that name too, but then thought that "Keep-Alive" might make us think like there's a watchdog sending no-op pulses every "keep-alive timeout" just to make sure the MergeService stays alive even without any activity. Like you'd send a keep-alive NO-OP to a service on a regular basis to keep it alive so it doesn't go to sleep. Which is actually quite the opposite of what we're willing to do here (killing the service if it's inactive, not keeping it alive).

But hey, naming is hard. Any suggestion welcome.

Sources/Bot/Services/MergeService.swift
AliSoftware: I created this doc-comment using Xcode auto-generation feature (Cmd-Alt-/ or something) so I just went with what it suggested 😄 
But yeah will try the alternate presentation, would make it nicer.

Tests/BotTests/MergeService/DispatchServiceContext.swift
AliSoftware: It's just to make sure we go _past_ it. We use the same idea already for all other timeouts in our tests, so I figured we should do the same here 🤷‍♂

ilyapuchka: dunno, feels weird to me

AliSoftware: agreed. updated to remove that multiplication

Sources/Bot/Services/DispatchService.swift
```diff
@@ -116,18 +122,27 @@ public final class DispatchService {
...
-                    logger.log("👋 MergeService for target branch `\(targetBranch)` is idle, destroying")
-                    self?.mergeServices.modify { dict in
-                        dict[targetBranch] = nil
-                    }
-                    self?.mergeServiceLifecycleObserver.send(value: .destroyed(service))
+            }
+        // Observe idle states to clean up dormant MergeServices only after they have been idle for too long
+        mergeService.state.producer
+            .filter { $0.status == .idle }
+            .debounce(self.idleMergeServiceCleanupDelay, on: scheduler)
```
AliSoftware, line 136: 🤔 just re-read my PR and realized, shouldn't those two lines be actually inverted? (be `debounce().filter()` instead of `filter().debounce()`)

If we `filter` first, there's not much to `debounce`, while what we want is to cancel the cleanup if the state goes out of `.idle` state

ilyapuchka: if what we want is to wait until state is `idle` and then delay it then I think the order is correct.

AliSoftware: Assuming the cleanupDelay is 5mn, I think if the state machine goes into `.idle` at T=0 for 2 minutes but then receives new PRs in its queue at T=2mn and thus goes back to `.ready` -> `.integrating` and stays there for 20mn… Then we don't want to kill the `MergeService` at T=5mn while it's in the `.integrating` state just because it's 5mn after the last `.idle`.

What we want instead is to only kill the service if it has remained in the `.idle` state for 5 minutes without going into a new state during those 5 minutes and thus been considered inactive for too long. Hence why I think the order of those lines might be wrong here.

Fixed it with a new commit; lmk if you think my logic is flawed here

ilyapuchka: > What we want instead is to only kill the service if it has remained in the .idle state for 5 minutes without going into a new state during those 5 minutes

I'm not sure if that's how debounce works but if we can validate it with tests then it's ok (and I'd expected such tests to fail with a different order)

AliSoftware: FYI I [wrote some unit tests](https://github.com/babylonhealth/Wall-E/pull/48/files#diff-8179eec3d972c5532226653636154701R231-R352) as you rightfully suggested, and it indeed passes with `.debounce().filter()` but fail with `.filter().debounce()`, confirming we should indeed debounce before filtering. 

* The latter (`filter().debounce()`) indeed destroys the `MergeService` too soon especially if it was in `.idle` at some point and then it switches to `runningStatusChecks` but stays there too long (the `filter(.idle).debounce(T)` would then destroy the `MergeService` at that point while that's not what we want)
* While the former (`debounce(T).filter(.idle)`) and now current implementation properly destroys the service ONLY if it doesn't leave the `.idle` state _at all_ for the given delay – so exactly what we want to achieve 👍

Sources/App/Extensions/EnvironmentProperties.swift
inamiy: I'm OK with current naming 👍

Sources/Bot/Services/DispatchService.swift
inamiy: Not sure why `debounce` is needed instead of `delay`. 
I think the correct operator sequence is `filter().delay()`.

AliSoftware: @inamiy the goal is to cancel the destruction if we go out of idle state before the end of the delay. if we use `delay` it won't be cancelled will it? While my understanding of debounce is that it will cancel the destruction of the service if an event make us go out of the idle state before the delay, right?

inamiy: @AliSoftware 
> to avoid constantly killing and recreating them

I missed this PR body.
As you mentioned, it's probably better to add comment to disambiguate the use of `debounce` from `delay`.

inamiy: @AliSoftware  https://github.com/babylonhealth/Wall-E/pull/48#discussion_r358983938
I now also agree with `debounce(T).filter(.idle)` is the correct approach.

inamiy (approved):

PR #51, opened 2020-01-23
[IOSP-543] Fix MergeBot deployment

AliSoftware:
## Why

One of the newly introduced env var was supposed to be optional, but was implemented as required, which made the last attempt at deploying the MergeBot fail (†)

Also another env var was providing a default value… but implemented as required nonetheless

_(†) We could just have added the env var to the manifest but making this one optional feels more logical, and will avoid failing for the same reason when Android decides to deploy their instance to latest tag too)_

## How

Made the 2 env vars optional, and not `throws` anymore, allowing the default value for it to be useful

Took the occasion to improve env vars doc

### DIFF
--- a/Sources/App/Extensions/EnvironmentProperties.swift
+++ b/Sources/App/Extensions/EnvironmentProperties.swift
@@ -32,13 +32,14 @@ extension Environment {
         return ["yes", "1", "true"].contains(stringValue.lowercased())
     }
 
-    /// Maximum time to wait for a status check to finish running and report a red/green status
+    /// Maximum time (in seconds) to wait for a status check to finish running and report a red/green status
     static func statusChecksTimeout() throws -> TimeInterval? {
         let value: String = try Environment.get("STATUS_CHECKS_TIMEOUT")
         return TimeInterval(value)
     }
 
-    /// Delay to wait after a MergeService is back in idle state before killing it. Defaults to 5 minutes.
+    /// Delay (in seconds) to wait after a MergeService is back in idle state before killing it.
+    /// Defaults to 300 seconds (5 minutes)
     static func idleMergeServiceCleanupDelay() throws -> TimeInterval? {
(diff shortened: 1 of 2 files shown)


Ben-Henshall (approved):

Sources/App/Extensions/EnvironmentProperties.swift
```diff
@@ -3,44 +3,53 @@ import Bot
...
+    /// If set to YES/1/TRUE, the Merge Bot will require all GitHub status checks to pass before accepting to merge
+    /// Otherwise, only the GitHub status checks that are marked as "required" in the GitHub settings to pass
     static func requiresAllGitHubStatusChecks() throws -> Bool {
         guard let stringValue: String = Environment.get("REQUIRES_ALL_STATUS_CHECKS") else {
             return false // defaults to only consider required checks
         }
         return ["yes", "1", "true"].contains(stringValue.lowercased())
     }
 
+    /// Maximum time to wait for a status check to finish running and report a red/green status
```
AliSoftware:
```suggestion
    /// Maximum time (in seconds) to wait for a status check to finish running and report a red/green status
```

Sources/App/Extensions/EnvironmentProperties.swift
```diff
@@ -3,44 +3,53 @@ import Bot
...
     static func statusChecksTimeout() throws -> TimeInterval? {
         let value: String = try Environment.get("STATUS_CHECKS_TIMEOUT")
         return TimeInterval(value)
     }
 
-    /// Delay to wait after a MergeService is back in idle state before destroying it
+    /// Delay to wait after a MergeService is back in idle state before killing it. Defaults to 5 minutes.
```
AliSoftware:
```suggestion
    /// Delay (in seconds) to wait after a MergeService is back in idle state before killing it.
    /// Defaults to 300 seconds (5 minutes)
```

danspndl (approved):
Looks good 🚀

ilyapuchka (approved):

PR #53, opened 2020-02-03
[IOSP-296] [IOSP-297] Logz.io-compatible JSON Logs

AliSoftware:
Tickets:

* https://babylonpartners.atlassian.net/browse/IOSP-296
* https://babylonpartners.atlassian.net/browse/IOSP-297

## Why?

We want our logs to be better formatted and processed on Logz.io

## How

Adopting the [JSON logging format supported by our Logz.io instance](https://engineering.ops.babylontech.co.uk/docs/observability/logzio/#logging-format), which parses JSON in logs to present it in a searchable way via Kibana & ElasticSearch

## Tested

Wrote some Unit Tests – which was probably more painful that is was worth because I had to create a new test target but still trying to share `TestUtils.swift` between the targets, and had a lot of linker issues with CCrypto libs and friends (and on top of that Xcode kept crashing when trying those changes… 🙀 ) 

I ended up dropping the idea of having a common TestUtils framework target, and just used Unix symlinks instead to share the file between both test targets. Kind of unorthodox but way simpler to manage and to make it done.


Package.swift
```diff
@@ -8,13 +8,14 @@ let package = Package(
         .package(url: "https://github.com/ReactiveCocoa/ReactiveSwift.git", from: "6.0.0"),
         .package(url: "https://github.com/Babylonhealth/ReactiveFeedback", from: "0.7.0"),
         .package(url: "https://github.com/Quick/Nimble.git", from: "8.0.2"),
-        .package(url: "https://github.com/krzyzanowskim/CryptoSwift.git", from: "1.3.0")
+        .package(url: "https://github.com/krzyzanowskim/CryptoSwift.git", from: "1.3.0"),
```
AliSoftware: Leftover of modifying the Package.swift while trying to add common test target framework…

```suggestion
        .package(url: "https://github.com/krzyzanowskim/CryptoSwift.git", from: "1.3.0")
```

Tests/AppTests/TestUtils.swift
```diff
@@ -0,0 +1 @@
+../TestUtils.swift
```
AliSoftware, line 1: symlink (see PR description)

Sources/App/configure.swift
```diff
@@ -11,15 +11,15 @@ enum ConfigurationError: Error {
 /// Called before your application initializes.
 public func configure(_ config: inout Config, _ env: inout Environment, _ services: inout Services) throws {
 
-    let logger = PrintLogger()
+    let logger = LogzIOLogger()
```
dmcrodrigues: Have we done this part?

> The log level should be also configurable through an environment variable to enable increasing/decreasing as needed without getting a new image. If not provided, INFO should be used as the default log level.

Sources/App/Logger/LogzIOLogger.swift
```diff
@@ -0,0 +1,64 @@
+import Foundation
+import Vapor
+
+public final class LogzIOLogger: Logger, Service {
+    let serializer = JSONEncoder()
+    public func log(_ string: String, at level: LogLevel,
```
dmcrodrigues:
```suggestion
    
    public func log(_ string: String, at level: LogLevel,
```

Sources/App/configure.swift
AliSoftware: Oh indeed missed that

Sources/App/Logger/LogzIOLogger.swift
```diff
@@ -0,0 +1,64 @@
+import Foundation
+import Vapor
+
+public final class LogzIOLogger: Logger, Service {
```
dmcrodrigues: Can we call this `JSONLogger` instead? Logz is just an implementation/service detail and the structure is very generic.

Tests/AppTests/Logger/LogzIOLoggerTests.swift
```diff
@@ -0,0 +1,54 @@
...
+            file: "somefile.swift",
+            function: "somefunction",
+            line: 1337,
+            column: 42
+        )
+
+        let serializer = JSONEncoder()
+        let data = try serializer.encode(message)
+
+        print(message)
```
dmcrodrigues: Do we need this print?

Tests/AppTests/Logger/LogzIOLoggerTests.swift
```diff
@@ -0,0 +1,54 @@
+import XCTest
+@testable import App
+
+class LogsTests: XCTestCase {
+    func test_logzio_json_formatting() throws {
```
dmcrodrigues: Good stuff in adding a test 👍

AliSoftware:
Found a bug while testing again, the ServiceFactory resolves the wrong logger for `RequestLoggerMiddleWare` at runtime, despite the logger being registered beforehand and the Middleware using `container.make(Logger.self)` 🤔 

It seems it was an issue already present before, making the MiddleWare use the default Logger (which is in fact `ConsoleLogger`) instead of our own, but before that PR, the fact that we accidentally used `ConsoleLogger` instead of `PrintLogger` was not that visible…

Will investigate further, have marked the PR as blocked until I find the fix.

dmcrodrigues (commented):
Just a couple of comments but looks great 👍

Tests/AppTests/Logger/LogzIOLoggerTests.swift
AliSoftware: Yeah, that's the whole goal of the Logger 😛 .

Any custom Logger can do whatever they want, like sending the logs to a server or print it to the console or whatnot. Here because the way our services are integrated with Logz.io in our infra is that Logz.io picks things that are logged to STDOUT, we need to put that string to stdout, hence the `print` 😉 without it nothing would get logged to Logz.io.

I'll add a comment

dmcrodrigues:
@AliSoftware we can register an instance of `RequestLoggerMiddleware` which can be initialised with our own logger directly.

`middlewares.use(RequestLoggerMiddleware(logger: logger))`

Sources/App/configure.swift
AliSoftware: @dmcrodrigues Thinking about it, is this AC still relevant though, given that we get our logs via Logz.io anyway whose whole point is to provide nice ways for filtering things?

I mean if you want to filter logs, all you have to do is add a filter in the Kibana tab of Logz.io, no need to redeploy not even to change an env var in the manifest as we can filter them live anyway…

Tests/AppTests/Logger/LogzIOLoggerTests.swift
AliSoftware: Fixed in 8953830

Sources/App/configure.swift
dmcrodrigues: I don't think there's a need to log all incoming requests, it's the only debug message that we have and why I think we should have info as the minimum log level.

AliSoftware: Ok. I'll address that tomorrow then 👍

dmcrodrigues: If it takes too much effort we can extract that part to a separate ticket, up to you.

Sources/App/Logger/LogzIOLogger.swift
AliSoftware: Fixed in e42eaa2

Sources/App/configure.swift
AliSoftware: Implemented in 8c1280c805902432da9d36c20a7c398fee3daca8 + 9faf4f0afeabdf14abbadb469737e97196e33fa1

Tests/AppTests/Logger/LogzIOLoggerTests.swift
AliSoftware: 🤦‍♂ I just saw that this comment was in the Tests code, not in the main implementation code… now your comment makes way more sense… removing this useless `print` now 😉

Sources/App/Extensions/EnvironmentProperties.swift
```diff
@@ -59,6 +59,24 @@ extension Environment {
...
+    ///  - `debug`
+    ///  - `info`
+    ///  - `warning`
+    ///  - `error`
+    ///  - `fatal`
+    /// Any log that is higher that the `minimumLogLevel` in this list will be filtered out. e.g. a `minimumLogLevel` of `info`
+    /// will filter out `verbose` and `debug` logs and will only print `info`,`warning`,`error` and `fatal` logs
+    static func minimumLogLevel() -> LogLevel {
+        guard let value: String = Environment.get("MINIMUM_LOG_LEVEL") else {
+            return .verbose // If env var is not set, log everything and don't filter out any log
```
dmcrodrigues: By default should be `info`, we shouldn't force consumers to provide a log level, verbose or even debug is simply too much logging by default.

AliSoftware: Yeah I quite hesitated on that. My take was: don't break existing behavior. So if you don't set the env var, print the same logs as before we had that filtering feature 🤷‍♂ 

I'm ok to set it to .info though, will push that after my meeting

Sources/App/Logger/LogLevel+Compare.swift
```diff
@@ -0,0 +1,33 @@
...
+
+    // Note: Didn't implement it as Comparable and `<` operator because `.custom` case doesn't follow semantics.
+    public func isAtLeast(minimumLevel: LogLevel) -> Bool {
+        // If minimumLevel is .custom, only keep the ones that are exactly using the same custom level name
+        if case .custom(let minValue) = minimumLevel, case .custom(let selfValue) = self {
+            return selfValue == minValue
+        }
+        // Otherwise, consider custom levels have the same "importance" as .info
+        let defaultIndex = LogLevel.allLevels.firstIndex(of: .info)!
+        let minIndex = LogLevel.allLevels.firstIndex(of: minimumLevel) ?? defaultIndex
```
dmcrodrigues: How can `minimumLevel` not exist in this context? Also assuming a default in here doesn't seem right anyway.

AliSoftware: As expressed by the comment, it's for for the case when `minimumLevel` is `.custom("Foo")`.

(`allLevels` might be a misleading name, as it doesn't contain all levels but all pre-defined levels. naming is hard.)

Sources/App/Logger/LogLevel+Compare.swift
```diff
@@ -0,0 +1,33 @@
...
+
+    /// From more verbose to less verbose (i.e. from less important to more important)
+    private static let allLevels: [LogLevel] = [
+        .verbose, .debug, .info, .warning, .error, .fatal
+    ]
+
+    // Note: Didn't implement it as Comparable and `<` operator because `.custom` case doesn't follow semantics.
+    public func isAtLeast(minimumLevel: LogLevel) -> Bool {
+        // If minimumLevel is .custom, only keep the ones that are exactly using the same custom level name
+        if case .custom(let minValue) = minimumLevel, case .custom(let selfValue) = self {
```
dmcrodrigues: I'm starting to think we shouldn't have imported `Logging`, this feels overly complex.

AliSoftware: Yeah, that's the only thing that `Logging` provide that we might not want. But the `custom` case only affects this Compare logic, so reimplementing everything ourselves and then provide a bridge between our own implementation and Vapor _just_ to avoid that `.custom` case felt quite overkill to me 😉

Sources/App/Extensions/EnvironmentProperties.swift
AliSoftware: Fixed in 99272de82626b7153aa3a8562e8db3466e5e73c1

Sources/App/Logger/LogLevel+Compare.swift
dmcrodrigues: > so reimplementing everything ourselves and then provide a bridge between our own implementation and Vapor

Can you elaborate? I don't understand the part _reimplementing_ when we already had a protocol to abstract the logger, which is what `Logging` does so the only thing missing was having log levels which could be easily done by us as specified in the ticket, instead of having this train of levels we could have just 3 and avoid all of this complexity. 

Honestly I feel like we jumped into the implementation without checking the tickets and the expected AC and consider the best option based on that. 

Another thing we are losing is the abstraction from Vapor, `Bot` should be a module completely independent of Vapor and it really isn't anymore even if it's just for the logging part, however, as I mentioned feels like we brought an unnecessary layer of complexity without a need for it.

Sources/Bot/Logging/Logger.swift
```diff
@@ -0,0 +1,48 @@
+import Foundation
+
+// MARK: LogLevel
+
+/// Different log levels
+public enum LogLevel: String, Equatable, CaseIterable {
```
dmcrodrigues, line 6: ❤️

dmcrodrigues (approved):

Tests/AppTests/Logger/JSONLoggerTests.swift
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    // Check that we print anything above but nothing below the .info level
+    func test_loglevel_compare_info_level() {
+        XCTAssertEqual(LogLevel.debug >= .info, false)
+        XCTAssertEqual(LogLevel.info  >= .info, true)
+        XCTAssertEqual(LogLevel.error >= .info, true)
+    }
+
+    // Check that we print anything above but nothing below the .info level
```
AliSoftware:
```suggestion
    // Check that we print only error logs when minimum level is .error
```

viorel15 (approved):

PR #54, opened 2020-02-12
[IOSP-569] Fix GitHub comments accidentally seen as PR reference

AliSoftware:
# Why?

Recently GitHub changed something in how they parse magic references in comments.

* If you have a `#` followed by a number, GitHub recognizes it as a Pull Request reference and automatically transforms it into a link to that Pull Request.
* Except that previously, if you wrapped that `#n` inside backticks –to make it pre-formatted like code–, GitHub ignored those. Which is why we used those backticks in the "Your request is `#n` in the queue" comments.
* The issue is that it seems they changed that recently and now they also recognize those `#n` as links to PRs _even inside backticks_

# How

Introduced a zero-width-space character after the `#` to make sure it won't be seen as a PR reference by GitHub

Took the occasion to also make the textual description of our queues cleaner and more readable, for when we access the bot's URL in a browser to check the queue states. (This doesn't affect the JSON representation that WallEView uses, only the representation in browser)

### DIFF
--- a/Sources/App/Logger/JSONLogger.swift
+++ b/Sources/App/Logger/JSONLogger.swift
@@ -52,6 +52,7 @@ extension JSONLogger {
             let df = DateFormatter()
             df.dateFormat = "yyyy-MM-dd'T'hh:mm:ss.SSSZ"
             df.locale = Locale(identifier: "en_US_POSIX")
+            df.timeZone = TimeZone(secondsFromGMT: 0)
             return df
         }()
 
--- a/Tests/BotTests/Canned Data/DispatchServiceQueueStates.swift
+++ b/Tests/BotTests/Canned Data/DispatchServiceQueueStates.swift
@@ -1,6 +1,6 @@
 import Foundation
 
-let DispatchServiceQueueStates: Data = """
+let DispatchServiceQueueStatesJSON: Data = """
 [
   {
     "targetBranch" : "branch1",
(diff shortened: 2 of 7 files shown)


Sources/App/Logger/JSONLogger.swift
```diff
@@ -52,6 +52,7 @@ extension JSONLogger {
             let df = DateFormatter()
             df.dateFormat = "yyyy-MM-dd'T'hh:mm:ss.SSSZ"
             df.locale = Locale(identifier: "en_US_POSIX")
+            df.timeZone = TimeZone(secondsFromGMT: 0)
```
AliSoftware, line 55: That's what happens when you implement and test something while visiting London last week: now that I'm back in France the test failed locally because of me being back in GMT+1 😛

Tests/BotTests/Canned Data/DispatchServiceQueueStates.swift
```diff
@@ -89,3 +89,21 @@ let DispatchServiceQueueStates: Data = """
...
 ]
 """.data(using: .utf8)!
+
+
+let DispatchServiceQueueStatesString = """
+## Merge Queue for target branch: branch1 ##
+
+State(
+ - status: integrating PR #1 (some-branch),
+ - queue:\(" ")
```
AliSoftware, line 99: had to use that `\(" ")` otherwise Xcode kept removing that trailing space from the string literal 🤦‍♂  (which made the test constantly fail, despite me trying to re-add the space and re-saving the file)

dmcrodrigues (approved):

danspndl (approved):
👏  LGTM 🚀

PR #52, opened 2020-01-29
[IOSP-515] [CNSMR-3231] Documentation

AliSoftware:
This documentation update is focused on being **general for the reusable OSS repo.**

A [separate PR is pending in our main ios repo](https://github.com/babylonhealth/babylon-ios/pull/10427) to document the _day-to-day usage_ of that bot at Babylon specifically.

### README Documentation

This covers:
 - the main idea of how the bot works for people interested in using it as an OSS-available bot
 - some light pointers about implementation details

I didn't think it was worth it to get too much into verbose details about implementation outside of the `DispatchService` and `MergeService` overview, as code for the rest is pretty standard Vapor, but figured that the states+feedback diagrams could be a really helpful visual aid at least.

### Diagrams

The diagrams were made using the well established [GraphViz `.dot` format](https://graphviz.gitlab.io/).

I used the https://edotor.net/ online editor to easily live-edit and render those diagrams, as commented at the top of the `.dot` files

### Other changes

I took the occasion to update the README part about the `WallEView` client app, fix a compilation issue with the project. I also moved the `WallEView.xcodeproj` project inside `WallEView/` so that it would not conflict with `xed .` nor `vapor xcode -y` anymore when working on the main bot project.

PR #55, opened 2020-02-14
[IOSP-569] Revert JSON Logger

AliSoftware:
## Why?

Recent builds 0.14.0 and 0.14.1 seems to correspond to our bot rebooting randomly.

Until we investigate that further via dedicated ticket https://babylonpartners.atlassian.net/browse/IOSP-578, we reverted to 0.13 yesterday

But we're still interested in applying the fix for `#n` in comments

## How?

* Cut a branch `release/0.13.1` from current tag `0.13.0`, in preparation for creating that hotfix 0.13.1 with those changes
* This PR then reapplies #52 (documentation) and #54 (`#n` in comments) – which are both merely about wording/copy updates without any behavior changes – in `release/0.13.1` so on top of `0.13.0`, but without the commit #53 about JSON logger that got included in `0.14.0`.

I also added an additional commit to temporarily disable the network request logs (6f03239) since they are super verbose and not really useful but without #53 we don't have `LogLevels` anymore to disable them yet. We could reintroduce them gated via `LogLevel` if we reintroduce #53 later

## Diff content

The diff on this PR should thus only contain the exact same changes than the ones introduced in #52 and #54, plus the `#if LOG_FULL_NETWORK_REQUESTS` added in 6f03239

(Only line not included from #54 in this PR is the line that took the occasion of fixing the timezone of the `JSONLogger`, since obviously `JSONLogger` doesn't exist in this 0.13.1 hotfix)


dmcrodrigues:
@AliSoftware why reverting this on `master`? We should create a release branch based on `0.13.0` to release `0.13.1` but `master` should remain unchanged.

dmcrodrigues:
> but since we don't want to have two diverging branches

this won't happen because the release branch will be deleted right after releasing `0.13.1`.

AliSoftware:
@dmcrodrigues see PR description "Branching vs revert commit": I felt like branching out would easily lead us to diverging states. But maybe a release branch is still better?

I can change it to a release branch if you prefer, I don't really mind (though as shown by the link provided in "PR diff Review" paragraph, it should result in the same)

AliSoftware:
Ok I'll branch out and force-push then 👍

dmcrodrigues:
It's only because 0.14.0 has been already released so it's a bit strange rolling back master to 0.13.0 to release 0.13.1 and then reapply all the changes all together when 0.14.0 already exists, seems a bit convoluted to follow that approach.

AliSoftware:
@dmcrodrigues branching done 👍

README.md
```diff
@@ -18,22 +18,125 @@ Pull Requests should meet a specific set of criteria before being merged.
...
+## How the bot works
 
-👷‍♀️ **WIP, come back later** 👷‍♂️
+Wall-E is a bot written in Swift using Vapor.
 
-![](https://media.giphy.com/media/26ybvJNaZZKpPONEc/giphy.gif)
+It works by listening to the GitHub webhook to know when labels are being added/removed and when Status Checks are updated on PRs
 
-### Client app
+### Detection of the Merge Label
```
dmcrodrigues, line 27: all this changes are expected to be here?

AliSoftware: Those are from #52 (Documentation Goal to document the MergeBot). This was a commit just on top of 0.13.0 (which didn't lead to any deployment since the only changes were `README.md` + some images used in that README added in `assets/`)

AliSoftware:
This 0.13.1 hotfix will not be needed after all since we decided to restore 0.14.1 (as pod restarts were apparently not due to a crash from 0.14.x)

Closing this in favor of https://github.com/babylonhealth/manifests/pull/21842

PR #57, opened 2020-03-31
[IOSP-590] Restore Queue order on bot reboot

AliSoftware:
# Why?

When the docker image / Kubernetes pod hosting our bot reboots, the queue was rebuilt from the list of labelled PRs, but the order of the queue was arbitrary

# How?

## Ordering of Pull Requests on boot

The bot now orders the Pull Requests by date at which the bot last commented about the PR being picked up.

* At startup, the bot fetches the list of all open Pull Requests, as it did before, and filtered them to only keep the ones with Merge label, to get the list of initial PRs to start the queue with. But instead of using that unordered list directly, now it also does the following:
* for each of those filtered PR, we fetch the list of GitHub comments, filter only for comments by the bot matching "Your pull request was accepted […]" body, and find the latest one of those. This gives us a date for when the PR was previously last added to the queue by the bot
* Once we have those clue dates for each PR, we order the list of PRs by those dates. If we didn't find a bot comment for a PR, we assume a date of `.distantFuture` which will make those PR go to the end of the list

## Bot comments

The bot now mention in the comment when a PR was picked because of a reboot (and the bot rebuilding its initial queue) or not (and the bot adding a PR because you just added the Merge label)

# Tested?

Added a dedicated unit test for it

### DIFF
--- a/Sources/Bot/API/Repository.swift
+++ b/Sources/Bot/API/Repository.swift
@@ -113,7 +113,7 @@ extension Repository {
                 default:
                     return .failure(.api(response))
                 }
-        }
+            }
         )
     }
 
--- a/Sources/Bot/API/Repository.swift
+++ b/Sources/Bot/API/Repository.swift
@@ -92,6 +92,14 @@ extension Repository {
         )
     }
 
+    func issueComments(in pullRequest: PullRequest) -> Resource<[IssueComment]> {
+        return Resource(
+            method: .GET,
+            path: path(for: "issues/\(pullRequest.number)/comments"),
+            decoder: decode
+        )
+    }
+
     func merge(head: PullRequest.Branch, into base: PullRequest.Branch) -> Resource<MergeResult> {
         return Resource(
             method: .POST,
@@ -114,6 +122,14 @@ extension Repository {
             method: .PUT,
             path: path(for: "pulls/\(pullRequest.number)/merge"),
             body: encode(MergePullRequestRequest(with: pullRequest)),
+            decoder: decode
+        )
+    }
+
+    var currentUser: Resource<GitHubUser> {
+        Resource(
+            method: .GET,
+            path: "user",
             decoder: decode
         )
     }
--- a/Sources/Bot/API/RepositoryAPI.swift
+++ b/Sources/Bot/API/RepositoryAPI.swift
@@ -49,8 +49,16 @@ public struct RepositoryAPI: GitHubAPIProtocol {
         return client.request(repository.publish(comment: comment, in: pullRequest))
     }
 
+    public func fetchIssueComments(in pullRequest: PullRequest) -> SignalProducer<[IssueComment], GitHubClient.Error> {
+        return client.request(repository.issueComments(in: pullRequest))
+    }
+
     public func removeLabel(_ label: PullRequest.Label, from pullRequest: PullRequest) -> SignalProducer<Void, GitHubClient.Error> {
         return client.request(repository.removeLabel(label: label, from: pullRequest))
+    }
+
+    public func fetchCurrentUser() -> SignalProducer<GitHubUser, GitHubClient.Error> {
+        return client.request(repository.currentUser)
     }
 }
 
--- a/Sources/Bot/Models/IssueComment.swift
+++ b/Sources/Bot/Models/IssueComment.swift
@@ -7,28 +7,9 @@ public struct IssueComment: Equatable {
 }
 
 extension IssueComment: Decodable {
-    private static let dateFormatter: DateFormatter = {
-        // 2020-03-30T11:26:18
-        let df = DateFormatter()
-        df.locale = Locale(identifier: "en_US_POSIX")
-        df.dateFormat = "yyyy-MM-dd'T'HH:mm:ssZ"
-        return df
-    }()
-
     private enum CodingKeys: String, CodingKey {
         case user
         case body
-        case created_at
-    }
-
-    public init(from decoder: Decoder) throws {
-        let container = try decoder.container(keyedBy: CodingKeys.self)
-        self.user = try container.decode(GitHubUser.self, forKey: .user)
-        self.body = try container.decode(String.self, forKey: .body)
-        let dateString = try container.decode(String.self, forKey: .created_at)
-        guard let date = Self.dateFormatter.date(from: dateString) else {
-            throw DecodingError.dataCorruptedError(forKey: .created_at, in: container, debugDescription: "Invalid date format")
-        }
-        self.creationDate = date
+        case creationDate = "created_at"
     }
 }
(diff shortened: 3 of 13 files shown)


Sources/Bot/API/Repository.swift
```diff
@@ -92,6 +92,14 @@ extension Repository {
         )
     }
 
+    func issueComments(in pullRequest: PullRequest) -> Resource<[IssueComment]> {
+        return Resource(
```
NSMyself, line 96:
```suggestion
        Resource(
```

Sources/Bot/API/RepositoryAPI.swift
```diff
@@ -49,9 +49,17 @@ public struct RepositoryAPI: GitHubAPIProtocol {
         return client.request(repository.publish(comment: comment, in: pullRequest))
     }
 
+    public func fetchIssueComments(in pullRequest: PullRequest) -> SignalProducer<[IssueComment], GitHubClient.Error> {
+        return client.request(repository.issueComments(in: pullRequest))
```
NSMyself, line 53:
```suggestion
        client.request(repository.issueComments(in: pullRequest))
```

Sources/Bot/API/RepositoryAPI.swift
```diff
@@ -49,9 +49,17 @@ public struct RepositoryAPI: GitHubAPIProtocol {
...
+    }
+
     public func removeLabel(_ label: PullRequest.Label, from pullRequest: PullRequest) -> SignalProducer<Void, GitHubClient.Error> {
         return client.request(repository.removeLabel(label: label, from: pullRequest))
```
NSMyself, line 57:
```suggestion
        client.request(repository.removeLabel(label: label, from: pullRequest))
```

Sources/Bot/API/RepositoryAPI.swift
```diff
@@ -49,9 +49,17 @@ public struct RepositoryAPI: GitHubAPIProtocol {
...
     }
+
+    public func fetchCurrentUser() -> SignalProducer<GitHubUser, GitHubClient.Error> {
+        return client.request(repository.currentUser)
```
NSMyself, line 61:
```suggestion
        client.request(repository.currentUser)
```

Sources/Bot/Services/MergeService.swift
```diff
@@ -356,11 +359,42 @@ extension MergeService {
...
+    fileprivate static func whenStarting(
+        initialPullRequests: [PullRequest],
+        gitHubAPI: GitHubAPIProtocol,
+        botUser: GitHubUser?,
+        scheduler: DateScheduler
+    ) -> Feedback<State, Event> {
+        typealias DatedPullRequest = (pr: PullRequest, date: Date)
+
+        func isBotMergeComment(_ comment: IssueComment) -> Bool {
+            return (botUser.map({ comment.user.id == $0.id }) ?? true) // If botUser nil, default to true
```
NSMyself, line 384:
```suggestion
            (botUser.map({ comment.user.id == $0.id }) ?? true) // If botUser nil, default to true
```

Sources/Bot/Services/MergeService.swift
```diff
@@ -356,11 +359,42 @@ extension MergeService {
...
+                && comment.body.hasPrefix(MergeService.acceptedCommentIntro)
+        }
+
+        func datedPullRequest(for pullRequest: PullRequest) -> SignalProducer<DatedPullRequest, Never> {
+            return gitHubAPI.fetchIssueComments(in: pullRequest)
```
NSMyself, line 389:
```suggestion
            gitHubAPI.fetchIssueComments(in: pullRequest)
```

Sources/Bot/Services/MergeService.swift
```diff
@@ -356,11 +359,42 @@ extension MergeService {
...
+                .flatMapError { _ in .value([]) }
+                .map { comments in
+                    let lastBotComment = comments
+                        .filter(isBotMergeComment)
+                        .max { $0.creationDate < $1.creationDate }
+                    return (pullRequest, lastBotComment?.creationDate ?? .distantFuture)
```
NSMyself: Would this work?

```suggestion
                        .filter(isBotMergeComment)
                        .map(\.creationDate)
                        .min()
                    return (pullRequest, lastBotComment ?? .distantFuture)
```

NSMyself (approved):
Just nits. Good work!

Sources/Bot/API/Repository.swift
AliSoftware: I'll check but I'm not entirely sure this is supported on the Swift version we're using on Linux, one sec

NSMyself: Ah, that makes sense. Feel free to disregard then.

Sources/Bot/Services/MergeService.swift
AliSoftware: The goal is to find the max (most recent) date. I _think_ `max(<)` is semantically "order by `<` aka from smallest to biggest, then take the max value" aka "get the biggest date value", while `min(date)` would give us the smallest (aka oldest) date…

Makes me doubt though, I'll double-check the expected meaning of the parameter to `max`

AliSoftware: Updated code to use `.map` (couldn't use KeyPath since bot is not yet on latest Swift version) + `.max()`

Also updated the test to add a trap with a bot comment in Jan in the later PR before the later comment in March

See 8b6a72d20ab8f5cb6cd6d6bf9300bb14fdb9e681

Sources/Bot/Models/IssueComment.swift
```diff
@@ -0,0 +1,34 @@
...
+        case body
+        case created_at
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.user = try container.decode(GitHubUser.self, forKey: .user)
+        self.body = try container.decode(String.self, forKey: .body)
+        let dateString = try container.decode(String.self, forKey: .created_at)
+        guard let date = Self.dateFormatter.date(from: dateString) else {
```
ilyapuchka: no way to use date decoding strategy?

ilyapuchka (approved):
good idea! 👏

Sources/Bot/Models/IssueComment.swift
AliSoftware: Oh forgot about that one, good point.
And just saw that there's actually already code for that in `Decoding.swift` so all that shouldn't even be needed at all indeed.

AliSoftware: Should be fixed in 33d971a3d08332d8f1e7465720593ffc6e78a067 👍

